PR #48, opened 2023-02-02
47 settings and security

### LINKED ISSUE #47: Settings are not secure and could use more organization
CL is a very secure site and we can do even better here. How? Well, it's almost impossible to add headers like CSP to an old site, but it's easy to do with a new one. 

I'm going to pull all of the settings from CL as I can, and also include django-csp to really tighten it down.

Along the way, I'm going to analyze the settings in Django and pull over any that make sense. Once that's done, I'll finally be ready to try to deploy this in k8s.

mlissner:
This fixes #47 by adding a number of new settings and by adding the django-csp package.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,9 +1,11 @@
+# security.py
 #ALLOWED_HOSTS="*"  # <-- Uncomment this in dev
+SECRET_KEY=""
 
-HOSTNAME=""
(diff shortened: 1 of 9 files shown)


ERosendo (approved):
Looks good.

PR #52, opened 2023-02-02
fix(docker compose): Change the POSTGRES_DB variable

ERosendo:
This PR changes the value of the POSTGRES_DB environment variable to define the name for the default database that is created when the Postgres service is started.

I noticed that after merging https://github.com/freelawproject/bigcases2/pull/50 the django app was failing to start and then I realized the default name of the DB changed.

I think this PR will make developers' life easier.


mlissner:
Darn, sorry about this.

Issue #35, opened 2023-01-28
Add redis and a task queue


mlissner:
I think we're going to need this:

 - If we can wait a few seconds from when we get a webhook event, the doc might become available, if it's popular. Right now we don't have a way to delay doing something.
 - We don't have a good way to store idempotency keys. We could make a DB table for them, but redis with an expiration date would be better.
 - Receiving webhook events from CL has to be fast. There's a short timeout. That means that we need to store the webhook event, and give it a 200 response. Then, asynchronously, we should do things.

We use celery for our task queue in Django, and I've never liked it very much. Four years ago I did a review of [`rq`](https://python-rq.org/), but it wasn't good enough and I had to abandon it. Since then, it has slowly matured, and I think it'll have all the things we need for this project while being a LOT simpler than Celery.

mlissner:
I marked this for the pre-launch milestone. I'd be happy to push it back, if we think we can launch without it, but I don't think we really can. Can we?

anseljh:
Can confirm that RQ is nice! We used it at Syntexys. For a complex pipeline
it can be insufficient, but I think for this it'll do just fine.

mlissner:
I think a lot of the "noes" and blanks in the RQ column are now yeses and knowns.

mlissner:
I think there are two parts to this, or maybe three:

1. We need redis set up in docker-compose and in prod (where we use Elasticache.

3. We need a worker set up in docker-compose and deployed in k8s.  

1. There's a Sentry integration for rq. We'll want that set up via the Sentry settings.

1. Once Sentry is configured properly (and rq installed properly), we should set up a URL that triggers a failing `rq` task. CL has this as `celery_fail`, but it's the same idea. 

When all of that is wired up properly, you can hit the URL to trigger the failing `rq` task. Django will load that URL, which enqueues an `rq` task. The worker will pick that up and it'll fail. Sentry will pick up the failure, and we'll know it worked top to bottom.

mlissner:
OK, sorry, that's four parts. :facepalm:

mlissner:
Something else to do (five parts):
 - [ ] Look at CL's settings and see which parts need updating to work with redis. Two that I noticed today are setting the cache to use Redis and setting up sessions to use it (we don't use sessions much yet, but we will!)

mlissner:
Sorry, one more thing to do (six parts):

 - [ ] Add redis to the health check endpoint

mlissner:
[...]

PR #53, opened 2023-02-02
feat(redis): Add Redis and django-rq

ERosendo:
This PR fixes https://github.com/freelawproject/bigcases2/issues/35 by adding a number of new settings, adding Redis support and the django-rq package.

- Redis and django-rq are added as dependencies.

- This PR creates a file to store Redis settings and one to store setting related to RQ.

- This PR adds Redis and the RQ worker to the docker compose file

- The dockerfile is updated to create an image for the RQ worker.

- The sentry,py file is updated to add the RQ integration

- This PR creates an endpoint to trigger a failing rq task and also adds the endpoints to access the RQ dashboard

- The .env.local file is updated to include variables related to Redis.


docker/django/Makefile
```diff
@@ -23,9 +26,11 @@ all: image
 
 development:
 	docker build --target web-dev -t $(REPO):$(WEB_DEV) --build-arg BUILD_ENV=dev --file docker/django/Dockerfile .
+	docker build --target rq -t $(REPO):$(RQ_DEV) --build-arg BUILD_ENV=dev --file docker/django/Dockerfile .
 
 image:
 	docker build --target web-prod -t $(REPO):$(DOCKER_TAG_PROD) -t $(REPO):$(WEB_PROD) --file docker/django/Dockerfile .
+	docker build --target celery -t $(REPO):$(RQ_TAG) -t $(REPO):$(RQ_LATEST) --file docker/django/Dockerfile .
```
mlissner, line 33: Looks like Celery crept in.

.env.example
```diff
@@ -32,3 +32,6 @@ DB_NAME=""
 DB_USER=""
 DB_PASSWORD=""
 DB_HOST=""
+
+REDIS_HOST=""
+REDIS_PORT=""
```
mlissner, line 37: Rather than put this at the bottom, let's keep the file in the order of the settings files themselves. So, this should go between pacer.py and sentry.py, and we should add a comment like this:

```
# redis.py
REDIS_HOST=""
REDIS_PORT=""
```

mlissner (changes requested):
A couple small comments, but this looks great otherwise.

.env.example
ERosendo: Got it!

mlissner (approved):

PR #55, opened 2023-02-02
chore(makefile): build RQ images using the makefile

ERosendo:


docker/django/Makefile
```diff
@@ -37,6 +37,7 @@ push: image
 	@if [ $(UNAME) = "x86_64" ]; then \
 	    echo "Architecture is OK. Pushing.";\
 	    docker push $(REPO):$(DOCKER_TAG_PROD);\
+		docker push $(REPO):$(RQ_TAG);\
```
mlissner, line 40: I think we should push RQ_LATEST here too.

mlissner (commented):
One small comment, then please merge.

docker/django/Makefile
ERosendo: You're right!

Issue #34, opened 2023-01-27
Set up auto-deploy on successful merge


mlissner:
Most of our projects do this, and it's a pleasure. We should be able to just copy this from CourtListener once #4 is done.

mlissner:
Fixed via #56 and #59

PR #56, opened 2023-02-03
feat(ci): Add auto-deploy!

mlissner:
We won't know how well this works until it's merged, but it's copied from CourtListener's with one change. The one change is that it doesn't do collectstatic. Otherwise, it's pretty much the same, and I think I've got the right permissions set up for it to gooooo


mlissner:
Fixes #34.

ERosendo (approved):
Looks good. I hope it works

Issue #27, opened 2023-01-26
Post to Mastodon


mlissner:
Before we get twitter working (#26), we should do Mastodon. Might as well give the folks fleeing twitter something nice, and it'll give us a safe place to mess around.

Like with Twitter, we won't have a budget at first, so we can use the CL links that upgrade to PACER links. E.g.:

https://www.courtlistener.com/docket/66696476/01208457368/shawn-musgrave-v-mark-warner/

One nice thing about Mastodon is that we can edit our posts (though I'm not sure for how long). We might be able to use that to do a post with the bad link above, then edit the post to add thumbnails and better links later (but that might be hard).

anseljh:
I don't think there is a time limit for editing.

https://docs.joinmastodon.org/methods/statuses/#edit

PR #72, opened 2023-02-13
feat(subscriptions): Post to mastodon

ERosendo:
This PR adds support to create posts in mastodon and fixes https://github.com/freelawproject/bigcases2/issues/27.

- This PR creates a new model called `Post` to store data related to the messages the bot has sent in different channels

- We add a new template to create posts in Mastodon

- The function called `process_filing_webhook_event` is updated to start creating posts in mastodon.
 
- We add a new field to the `subscription` model. This new field is called `cl_slug` and We use it to create the link to the document. We also register the `subscription` model to the admin site so We can add the `cl_slug` to those subscriptions that were created before this PR.

- The `lookup` command is updated to get the `slug` from the API response and add it to the subscriptions when the --add flag is used.

### DIFF
--- a/bc/subscription/tasks.py
+++ b/bc/subscription/tasks.py
@@ -1,3 +1,5 @@
+from django.db import transaction
+
 from bc.channel.models import Post
 from bc.channel.selectors import get_mastodon_channel
 from bc.channel.utils.masto import post_status as mastodon_post
@@ -6,6 +8,7 @@ from bc.core.utils.messages import MASTODON_POST_TEMPLATE
 from .models import FilingWebhookEvent, Subscription
 
 
+@transaction.atomic
 def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
     """Process an event from a CL webhook.
 
@@ -18,15 +21,17 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
         return filing_webhook_event
 
     try:
-        subscription = Subscription.objects.get(
-            cl_docket_id=filing_webhook_event.docket_id
-        )
+        with transaction.atomic():
+            subscription = Subscription.objects.get(
+                cl_docket_id=filing_webhook_event.docket_id
+            )
     except Subscription.DoesNotExist:
         # We don't know why we got this webhook event. Ignore it.
         filing_webhook_event.status = FilingWebhookEvent.FAILED
         filing_webhook_event.save()
         return filing_webhook_event
 
+    filing_webhook_event.status = FilingWebhookEvent.SUCCESSFUL
     filing_webhook_event.subscription = subscription
     filing_webhook_event.save()
 


bc/core/utils/messages.py
```diff
@@ -0,0 +1,3 @@
+NEW_FILING_TEMPLATE = (
+    "New filing in {docket}: {desc} \n\nDoc. on CourtListener: {link}"
+)
```
mlissner, line 3: Since mastodon lets us have longer posts, a few things:

1. Let's include a link to the docket.
2. Let's use the link that redirects to the PDF (I think people prefer the PDF).
3. I tweaked the variable name

So:

```suggestion
MASTODON_POST_TEMPLATE = (
    """New filing in {docket}: {desc}

PDF: {pdf_link}
Docket: {docket_link}"""
)
```

mlissner (changes requested):
OK, this looks pretty good, but please see my smattering of suggestions and requested changes.

Let me know what you think!

bc/subscription/models.py
```diff
@@ -59,20 +66,20 @@ class Subscription(AbstractDateTimeModel):
         blank=True,
     )
 
-    def cl_url(self):
+    def cl_url(self) -> str:
         return f"https://www.courtlistener.com/recap/gov.uscourts.{self.cl_court_id}.{self.pacer_case_id}"
 
-    def pacer_district_url(self, path):
+    def pacer_district_url(self, path) -> Optional[str]:
```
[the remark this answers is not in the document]
ERosendo: You're right! It's nicer and simpler. I will add [pyupgrade](https://github.com/asottile/pyupgrade) as a pre-commit hook to automatically upgrade syntax for newer versions of the language.

bc/subscription/models.py
```diff
@@ -143,9 +154,61 @@ class FilingWebhookEvent(AbstractDateTimeModel):
...
+        blank=True,
+    )
 
     class Meta:
         indexes = [
             models.Index(fields=["docket_id"]),
             models.Index(fields=["pacer_doc_id"]),
         ]
+
+    def document_link(self) -> Optional[str]:
```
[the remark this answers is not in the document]
ERosendo: Got it! I will add the new properties!

bc/subscription/models.py
```diff
@@ -143,9 +154,61 @@ class FilingWebhookEvent(AbstractDateTimeModel):
...
+        if self.attachment_number:
+            return (
+                f"Doc {self.document_number}-{self.attachment_number} "
+                f"from {self.description}"
+            )
+
+        return f"Doc {self.document_number} " f"from {self.description}"
+
+
+class Post(AbstractDateTimeModel):
```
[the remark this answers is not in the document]
ERosendo: You're right! the subscription app should handle stuff related to CL and the cases the bot is following. It makes sense to move this model to the channel app. 

Thanks for your suggestion.

bc/subscription/models.py
```diff
@@ -143,9 +154,61 @@ class FilingWebhookEvent(AbstractDateTimeModel):
...
+    filing_webhook_event = models.ForeignKey(
+        "FilingWebhookEvent", related_name="posts", on_delete=models.CASCADE
+    )
+    channel = models.ForeignKey(
+        "channel.Channel", related_name="posts", on_delete=models.CASCADE
+    )
+    object_id = models.PositiveBigIntegerField(
+        help_text="The object's id returned by the channel API",
+    )
+
```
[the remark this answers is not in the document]
ERosendo: You're right!

bc/subscription/models.py
```diff
@@ -143,9 +154,61 @@ class FilingWebhookEvent(AbstractDateTimeModel):
...
+class Post(AbstractDateTimeModel):
+    filing_webhook_event = models.ForeignKey(
+        "FilingWebhookEvent", related_name="posts", on_delete=models.CASCADE
+    )
+    channel = models.ForeignKey(
+        "channel.Channel", related_name="posts", on_delete=models.CASCADE
+    )
+    object_id = models.PositiveBigIntegerField(
+        help_text="The object's id returned by the channel API",
+    )
```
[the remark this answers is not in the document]
ERosendo: I think It's a great idea.

bc/core/utils/messages.py
ERosendo: This is how the new template looks:

image

bc/subscription/services.py
```diff
@@ -19,11 +19,13 @@ def create_subscription_from_docket(docket):
 
     cl_docket_id = docket["id"]
     cl_court_id = docket["court_id"]
+    cl_slug = docket["slug"]
 
     pacer_court_id = docket["court_id"]
 
     court = find_court_by_id(cl_court_id)
     court_name = court[0]["name"] if len(court) == 1 else ""
+    pacer_court_id = court[0]["id"] if len(court) == 1 else docket["court_id"]
```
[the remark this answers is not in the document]
ERosendo: Got it! I will remove the pacer_court_id from the model and use the helper functions instead.

bc/subscription/models.py
```diff
@@ -187,6 +187,16 @@ def document_link(self) -> Optional[str]:
...
 
+    @property
+    def cl_pdf_or_pacer_url(self) -> str:
+        return f"{self.cl_document_url()}?redirect_to_download=True"
+
+    @property
+    def cl_docket_url(self) -> str | None:
+        if not self.subscription:
+            return None
+        return self.subscription.cl_url()
```
mlissner, line 198: Shouldn't this be a property for consistency?

bc/subscription/tasks.py
```diff
@@ -8,15 +13,34 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
     :return: A FilingWebhookEvent object that was updated.
     """
     filing_webhook_event = FilingWebhookEvent.objects.get(pk=fwe_pk)
-    try:
-        if filing_webhook_event.docket_id:
+    if filing_webhook_event.docket_id:
```
mlissner, line 16: Rather than doing a 30-line indent, you should do an early exit from the function here. Flip the logic, so you have:

```suggestion
    if not filing_webhook_event.docket_id:
        return filing_webhook_event
    
    # Then the rest can be outdented here
```

bc/subscription/tasks.py
```diff
@@ -8,15 +13,34 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
...
+        try:
             subscription = Subscription.objects.get(
                 cl_docket_id=filing_webhook_event.docket_id
             )
-            filing_webhook_event.subscription = subscription
+        except Subscription.DoesNotExist:
+            filing_webhook_event.status = FilingWebhookEvent.FAILED
```
mlissner, line 22:
```suggestion
            # We don't know why we got this webhook event. Ignore it.
            filing_webhook_event.status = FilingWebhookEvent.FAILED
```

bc/subscription/tasks.py
```diff
@@ -8,15 +13,34 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
     :return: A FilingWebhookEvent object that was updated.
     """
     filing_webhook_event = FilingWebhookEvent.objects.get(pk=fwe_pk)
```
mlissner, line 15: This all looks good. One last thing is that you'll want it wrapped in a transaction.

mlissner (changes requested):
Back at you with a few little things, but I think this will probably be everything. Thanks!

bc/subscription/models.py
ERosendo: You're right!

mlissner (approved):

PR #84, opened 2023-02-16
feat(mastodon): Create posts using long descriptions

ERosendo:
This PR adds long descriptions to mastodon posts.

- Adds a dataclass to create templates. 
        
    - The current version of the bot uses just a `string` for the template but I think We can do better using a `class` because We can store a few **attributes** related to the template like the **number of links** and the **number of characters** in the template, We can also create methods and use template's attributes,

- Adds a utility function to truncate long strings

- Renames the field called `description` in the `FilingWebhookEvent` model to  `short_description`

- Adds a new field to the `FilingWebhookEvent` model to store long descriptions.

- Updates the `process_filing_webhook_event` method to use long descriptions when we lack the short one.


bc/core/utils/messages.py
```diff
@@ -1,4 +1,26 @@
...
+
+POST_TEMPLATE = MastodonTemplate(
+    number_of_links=2,
+    character_count=32,
+    str_template="""New filing in {docket}: {desc}
 
 PDF: {pdf_link}
-Docket: {docket_link}"""
+Docket: {docket_link}""",
+)
```
mlissner, line 26: Can we add the doc number here too, if there is one? Perhaps:

> New filing in {docket}
> Doc #xyz: {desc}
> 
> PDF: 
> Docket:

bc/subscription/tasks.py
```diff
@@ -35,9 +36,21 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
...
+        description = filing_webhook_event.short_description
+    else:
+        docket_length = len(subscription.docket_name)
+        available_space = 500 - len(POST_TEMPLATE) - docket_length
+        if len(filing_webhook_event.long_description) > available_space:
+            description = trunc(
+                filing_webhook_event.long_description, available_space, "...👇"
+            )
+        else:
+            description = filing_webhook_event.long_description
```
mlissner, line 49: I think all of this should be in the Mastodon template as a method. It's a lot of detailed stuff inside a task that should really just be focused on sending posts and tweets.

We should also store the max Mastodon length in a variable we import wherever we use it. Who knows if it'll get longer.

bc/subscription/tasks.py
```diff
@@ -35,9 +36,21 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
...
 
-    message = MASTODON_POST_TEMPLATE.format(
+    if filing_webhook_event.short_description:
+        description = filing_webhook_event.short_description
+    else:
+        docket_length = len(subscription.docket_name)
+        available_space = 500 - len(POST_TEMPLATE) - docket_length
+        if len(filing_webhook_event.long_description) > available_space:
+            description = trunc(
+                filing_webhook_event.long_description, available_space, "...👇"
```
mlissner, line 46: There's an ellipsis unicode character that looks better and only takes one byte from the length of our posts: `…`. But also, we shouldn't do this until we have the photo part ready.

mlissner (changes requested):
A few tweaks, but nothing major. Thank you!

bc/core/utils/messages.py
ERosendo: Sure!

ERosendo: The posts using the new template look like this:

image

bc/subscription/tasks.py
ERosendo: got it!

bc/subscription/tasks.py
ERosendo: I agree. I think We can add another property on the `FilingWebhookEvent` model to get the description, something like:

```python 
@property
def description(self) -> str:
        if self.short_description:
            return self.short_description
        return self.long_description
```

and We can also add a method in the Mastodon template to truncate the string if needed

mlissner (approved):

mlissner:
Seems overly complicated, but looks like it'll work, so OK. We're going to need to start writing more tests pretty soon here.

PR #85, opened 2023-02-16
fix(subscriptions):  Allow numberless entries and add minute template

ERosendo:
This PR adds support to allow numberless entries and adds a new template to create posts in Mastodon using the Webhook events. This PR fixes https://github.com/freelawproject/bigcases2/issues/79.

### DIFF
--- a/bc/core/utils/messages.py
+++ b/bc/core/utils/messages.py
@@ -3,9 +3,15 @@ from dataclasses import dataclass
 from bc.core.utils.string_utils import trunc
 
 
+class AlwaysBlankValueDict(dict):
+    """Just return blank, regardless of the key"""
+
+    def __getitem__(self, key):
+        return ""
+
+
 @dataclass
 class MastodonTemplate:
-    character_count: int
     str_template: str
     link_placeholders: list[str]
     max_characters: int = 500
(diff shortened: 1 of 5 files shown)


bc/core/utils/messages.py
```diff
@@ -57,3 +57,10 @@ def format(self, *args, **kwargs) -> str:
 PDF: {pdf_link}
 Docket: {docket_link}""",
 )
+
+
+MINUTE_TEMPLATE = MastodonTemplate(
+    character_count=22,
```
mlissner, line 63: I guess I missed this the first time, but having magic numbers is usually bad. Is 23 coming from the length of the template itself? Is there a way we could compute this instead?

mlissner (changes requested):
I'm surprised this works! Could you do a couple tests to prove that the various templates work properly, especially with edge cases?

bc/core/utils/messages.py
ERosendo: The `character count` is the number of fixed characters of the template. For example: 

- The template string for the **minute entries** is `"New minute entry in {docket}: {description}"`. This template has 2 placeholders ( `docket` and `description` ). We need to remove those placeholders and compute the length of the new string to get the `character count`. In this case, the string without the placeholders is: `New minute entry in : `

I think We can compute this value using a regular expressions that removes elements encloses by `{}`. We can use the following expression:

```python
clean_template = re.sub("\{.*?\}", '' ,str_template)
return len(clean_template)
```

mlissner: We discussed this on Slack, but I thought I'd copy my solution to this puzzle here in case others are following our work: 

```
t = 'for {bar}'
class AlwaysBlankValuesDict(dict):
    """Just return blank, regardless of the key"""
    def __getitem__(self, key):
        return ""
len(t.format_map(BlankKeyDict()))  # returns 4
```

`AlwaysBlankValuesDict` returns `""` no matter what key you ask it for, so it lets you take a template and put blanks in all the placeholders, then measure its length.

mlissner (approved):

Issue #82, opened 2023-02-16
Don't broadcast junk docket entries


mlissner:
BCB1 has a little regex to identify posts that nobody cares about, like admission post hoc, blah. Let's grab that an integrate it into BCB2.

PR #86, opened 2023-02-17
feat(core): Add a regex expression to avoid posting junk docket entries.

ERosendo:
This PR fixes https://github.com/freelawproject/bigcases2/issues/82.

### DIFF
--- a/bc/core/utils/messages.py
+++ b/bc/core/utils/messages.py
@@ -1,7 +1,9 @@
+import re
 from dataclasses import dataclass
 
 from bc.core.utils.string_utils import trunc
 
+DO_NOT_POST = re.compile('(pro hac vice|notice of appearance|certificate of disclosure|corporate disclosure|add and terminate attorneys)', re.IGNORECASE)
 
 class AlwaysBlankValueDict(dict):
     """Just return blank, regardless of the key"""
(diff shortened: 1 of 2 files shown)


bc/subscription/tasks.py
```diff
@@ -35,28 +35,29 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
...
-    channel = get_mastodon_channel()
-    Post.objects.create(
-        filing_webhook_event=filing_webhook_event,
-        channel=channel,
-        object_id=api_post_id,
-        text=message,
-    )
-
-    return filing_webhook_event
+    if not DO_NOT_POST.search(filing_webhook_event.description):
```
mlissner, line 38: Flip your logic please to avoid indentation. :)

bc/core/utils/messages.py
```diff
@@ -1,7 +1,13 @@
+import re
 from dataclasses import dataclass
 
 from bc.core.utils.string_utils import trunc
 
+DO_NOT_POST = re.compile(
+    "(pro hac vice|notice of appearance|certificate of disclosure|corporate disclosure|add and terminate attorneys)",
+    re.IGNORECASE,
+)
```
mlissner, line 9: I think re.VERBOSE might be nice here, but it's not a must. Wouldn't have mentioned it except for the other comment I wound up making.

mlissner (changes requested):
Two little ones, thanks.

bc/subscription/tasks.py
ERosendo: done! I realized I can add an early return to avoid the indentation

mlissner (approved):

PR #89, opened 2023-02-17
feat(core): Add docket link in the template for minute entries

ERosendo:
This PR adds the docket link to the `MINUTE_TEMPLATE`.

### DIFF
--- a/bc/core/utils/messages.py
+++ b/bc/core/utils/messages.py
@@ -87,6 +87,8 @@ Docket: {docket_link}""",
 
 
 MINUTE_TEMPLATE = MastodonTemplate(
-    link_placeholders=[],
-    str_template="""New minute entry in {docket}: {description}""",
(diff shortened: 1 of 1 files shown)


bc/core/utils/messages.py
```diff
@@ -68,6 +68,7 @@ def _available_space(self, *args, **kwargs) -> int:
     def format(self, *args, **kwargs) -> str:
         if "description" in kwargs:
             available_space = self._available_space("description", **kwargs)
+            print(available_space)
```
mlissner, line 71: Oops.

mlissner: Can you fix with an amended commit and a force-push so our commit log isn't too polluted?

ERosendo: Sure!

mlissner:
Nice to see that we didn't have to count any characters to make this tweak. :)

Issue #88, opened 2023-02-17
Long case titles are a problem


mlissner:
[This case](https://www.courtlistener.com/docket/65636836/application-of-the-new-york-times-company-and-charlie-savage-for-access-to/) is a problem because its title is:

> APPLICATION OF THE NEW YORK TIMES COMPANY AND CHARLIE SAVAGE FOR ACCESS TO CERTAIN DOCKETS, ORDERS, LEGAL BRIEFING, AND ARGUMENT TRANSCRIPTS ANCILLARY TO JANUARY 6 GRAND JURY PROCEEDINGS

That's most of a tweet. The simple way to fix this is to allow people to tweak the case name when using the `lookup` command.

PR #90, opened 2023-02-17
feat(subscription): Add new argument to the lookup command

ERosendo:
This PR fixes https://github.com/freelawproject/bigcases2/issues/88 by adding a new argument called `name` to the `lookup` command

### DIFF
--- a/bc/subscription/management/commands/lookup.py
+++ b/bc/subscription/management/commands/lookup.py
@@ -16,6 +16,12 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument("cl-id", type=str)
         parser.add_argument(
+            "--name",
+            type=str,
+            help="Custom name for the case. This argument overwrites the name from CL",
+            required=False,
+        )
+        parser.add_argument(
             "--add",
             action="store_true",
             help="Save and Subscribe to the case found using the CL API",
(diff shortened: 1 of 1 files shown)


mlissner:
This approach won't work because you won't necessarily know it's a really long name until after the case has been added. It should prompt you for the name instead, with the default set to whatever comes back from the API.

ERosendo:
Got It! 

Will the command always prompt you for the name?

We can set a `max number of characters`, so the command will only prompt for the name when We get a string longer than the `max_number`.

let me know what you think.

mlissner:
I would say it should just prompt you every time, but have a default, so pressing enter is all you have to do.

ERosendo:
Cool. I'm going to refactor the command to that.

ERosendo:
Here's some gifs showing the command:

for short names:
short-string-command

for long names:
long-string-command

Let me know what you think.

mlissner:
You're close, but I'll be picky. The way it should work is it should prompt like this:

> Name (Hillary v. Trump): ▉

If you press enter it just does what's in the parens. If you don't, then you can type what you want.

ERosendo:
I know that including the default value in the prompt between brackets is a fairly common convention but I tried that approach and I think It doesn't look clean when the default is a long string. Here's a gif showing the prompt using brackets:


long-string-command-brackets


It looks good for short string, though!

Issue #81, opened 2023-02-16
Sometimes we only have the long description, not the short, so we need to post the long.


mlissner:
In BCB1, it only got PACER data from RSS feeds. That worked pretty well, but it could be slow and not every jurisdiction has an RSS feed. 

We can do better now, since we're using all the sources in RECAP, but for popular cases, this can mean that we get the long description (the entry description) before we get the short description. 

We need to handle this in the bot, a few solutions are:

1. Wait for the RSS feed, then use the short description from there. This sucks, and anyway, it wouldn't work very well because some jurisdictions lack RSS feeds.

3. Purchase the missing short description. We could do this, but it costs money, and for what? A lower fidelity thing? Nah.

5. Use the long description somehow. 

I'm partial to option 3. So, when we get the long description and lack the short description, what we should do is:

1. Post as much of the short description as the posting limit allows on each platform (twitter/mastodon, etc.)
2. If there is more than can't be posted, we:
    3.  truncate the post (see [`trunc`](https://github.com/freelawproject/courtlistener/blob/c029929f6285833f9c22ed4a95211e3d01f712b2/cl/lib/string_utils.py#L5-L33) for a better truncate method than what Python provides).
    4. Do a little down hand emoji at the end of the truncation.
    5. Put the full description in the first screenshot. (We'll need a Python library for this; I'm not sure what exists, but simple would be nice!)

I think it'd be smart to deliver this in two PRs. First, do step 1 so users get _something_. Then do step 2, with the sreenshot, since that will be trickier and take time.

@ERosendo, we also discussed tweaking the template so the description is last, but I think we shouldn't do that. We should keep it in the order it's in now with the links last.

PR #91, opened 2023-02-23
feat(channels): Add text images to posts in mastodon. 

ERosendo:
This PR adds text images to posts that use long descriptions and closes https://github.com/freelawproject/bigcases2/issues/81. 

- Adds Pillow as a dependency 

- Adds a down hand emoji at the end of the truncated string.

- Adds a new class called `TextImage`. This new class has all the methods to create a 16:9 image that shows the title of the case and the long description.

- Updates the `format `method of the `MastodonTemplate` to create the images when a long description is used. 

- Adds a new parameter to the `post_status` method in the `maston.py` file. This new parameter is called text_image and allows us to create a post with an image attached. 

Here's one of the text pictures created with this new class:

image

This is how the post looks on Mastodon:

image


mlissner:
Can you please request my review again once tests have passed and it's not marked as a draft anymore?

bc/core/utils/images.py
```diff
@@ -0,0 +1,273 @@
...
+    title: str
+    description: str
+    title_font_path: Path = (
+        STATIC_GLOBAL_ROOT / "fonts" / "CooperHewitt-Bold.otf"
+    )
+    desc_font_path: Path = (
+        STATIC_GLOBAL_ROOT / "fonts" / "CooperHewitt-Light.otf"
+    )
+    font_size: int = 24
+    font_spacing: int = 16
```
mlissner, line 24: Font spacing would normally mean the distance between letters, also known as the kerning. What you mean here is line height, right? If so:

```suggestion
    line_height: int = 16
```

Or perhaps:

```suggestion
    line_spacing: int = 16
```

bc/core/utils/images.py
```diff
@@ -0,0 +1,273 @@
...
+        number of pixels in a 16:9 image. This method does not include padding in the
+        equation to keep it simple.
+
+        Args:
+            length (int): Number of the pixels of the text when rendered.
+
+        Returns:
+            int: height (in pixels)
+        """
+        return ceil(sqrt((9 / 16) * (self.line_height) * length))
```
mlissner, line 81: Well, I might be able to make this make sense if I stare at it awhile longer, but is there something that you can do to make it intuitive why this works? 

Also, please remove extra parens:

```suggestion
        return ceil(sqrt((9 / 16) * self.line_height * length))
```

mlissner (changes requested):
Well, making an image is certainly harder than I thought it would be. A couple thoughts:

1. Whole dissertations have been written on line wrapping, and people get really passionate about doing it well. Glad we didn't get sucked into that rabbit hole!

2. I think your approach is really good, but I worry about how it performs on really long text. Could we somehow set a max width and then start making the image taller if it needs to go beyond that? I don't want to wind up with a 16×9 image with text so small nobody can read it on a reasonably-sized screen. Your approach here could have that problem, right?

bc/core/utils/images.py
ERosendo: You're right about the `font_spacing`.

I like your second suggestion because We're using this `attribute` to control the `spacing` parameter of the [multiline_text](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.multiline_text) method from the [ImageDraw ](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html) module and they use the following description in the docs: 

> spacing – The number of pixels between lines.

It's really vague but I think they mean that this parameter sets **the vertical gap** between lines.

I don't think We should call it `line_height` because We're already using that name to represent the combined measurement of the `font_size` and this space between lines. 

I thought about calling this attribute **leading** but It seemed too technical.

bc/channel/utils/masto.py
```diff
@@ -57,8 +57,13 @@ def subscribe(force=False):
     return response
 
 
-def post_status(message: str) -> int:
+def post_status(message: str, text_image: bytes | None) -> int:
     m = get_mastodon()
-    api_response = m.status_post(message)
+    media_ids = None
+    if text_image:
+        media_dict = m.media_post(text_image, mime_type="image/png")
```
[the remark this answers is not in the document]
ERosendo: alright! I'm going to add the `description` and the `focus`. 

Thanks for your comment.

bc/core/utils/images.py
ERosendo: @mlissner let's suppose We have a **rectangle** with height **H** and width **W** and this **rectangle** has a **16:9** aspect ratio (Mastodon crops the images using that ratio), then the following  is always `true`:

``` 
 1) W/H = 16/9
```

If We want to render a paragraph inside this **rectangle**, We need to follow these rules:

``` 
 a) the width of each line should be W pixels at most ( this avoids horizontal overflow)
 b) the height of the paragraph (after wrapping the text) should be less than H ( This avoids vertical overflow)
```

We can write **b** like the following inequation:

```
 2) H ≥ paragraph_height 
```

We can approximate the value of the **paragraph_height** using the **number of lines** and the **line height**, so:

```
3) paragraph_height ≈ line_count * line_height
```
Now, We can use the expression from (3) in (2). Hence, inequation (2) becomes

```
4) H ≥ line_count * line_height
```

We still need the **line_count** but We can compute the width (in pixels) of the text without wrapping (the text is rendered as a single line) and then We can use rule **a** to get an estimate of the number of lines.  Hence, We can express **line_count** like:

```
5) line_count ≈ ( width_single_line / W )
```

If we use (5) in inequation (4), We got the following:

```
6)  H ≥ ( width_single_line / W ) * line_height 

7)  H ≥ ( width_single_line * line_height)/ W 
```

We can use (1) and solve for **W**. Hence, equation (7) becomes:

```
8) H ≥ ( width_single_line * line_height)/ [ (16/9) * H ]

9) H ≥ ( width_single_line * line_height ) / ( 16 / 9)

10) H ≥ √ [ (9/16) * ( width_single_line * line_height) ]
```

We're using inequation 10 to get the initial dimensions of the rectangle ( We didn't include padding in the equation to keep it simple). 

Let me know what you think.

bc/settings/django.py
```diff
@@ -15,7 +15,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", default=True)
-
+STATIC_GLOBAL_ROOT = BASE_DIR / "bc/assets/static-global/"
```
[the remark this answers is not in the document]
ERosendo: You're right! I will use the existing static file settings.

ERosendo:
You're right about the issue with really long descriptions. This approach creates a 16:9 image but the text is really small. Here's an example of a post with a long description

This is how It looks using chrome on a 1080p monitor ( the resolution is 1920x1080):

| 1080p Monitor  | Mobile device |
| ------------- | ------------- |
|   |   |



We can use your idea about setting a max width and start making the image taller. I've been doing some tests and I think We should set the max width to **854px**, Images with that width looks fine on mobile screens. Here's one of them:




Let me know what you think.

mlissner:
Yeah, that max-width sounds good to me. Thanks for checking it out.

bc/core/utils/images.py
mlissner: Goodness, this was quite a puzzle. I think we either need this explanation in the code or a reference to it. My hope is that future tweaks to the code won't have to figure this out all over again.

ERosendo:
@mlissner I've added the `max-width` attribute to the `TextImage` class. I changed the default value to 700px because I think it looks better on mobile screens.

mlissner (changes requested):
I think this looks good. The only changes I'd like to see are:

1. Remove the collectstatic if possible (as discussed in Slack).

3. Either add a reference to the comment about the math to the code or add the comments themselves to the code. In six months, when we're tweaking this, we'll be glad not to have to hunt for your notes in this PR!

mlissner (approved):

mlissner:
This is great, thank you!

PR #96, opened 2023-02-27
Make too-long posts shorter

mlissner:
This is a second pass at the work in https://github.com/freelawproject/bigcases2/pull/94. 

Looking at this post: https://law.builders/@bigcases/109938980184504362

I think it should be shorter before going to the "See more" so this does that. I think if we're going to provide a picture, then we shouldn't provide so much text.


ERosendo (approved):
LGTM

PR #98, opened 2023-02-28
feat(setting): Tweak static file settings

ERosendo:
This PR adds settings to use **S3** and **CloudFront** to serve static files.


mlissner (approved):
I'm assuming this is all coming more or less from CL. It looks good to me. Let's see if it works!

PR #99, opened 2023-03-01
feat(web): Add Tailwind CSS

ERosendo:
This PR adds Tailwind CSS to the `web` app using the [django-tailwind](https://github.com/timonweb/django-tailwind) package

- Adds the the [django-tailwind](https://github.com/timonweb/django-tailwind) package as a dependency

- Updates the dockerfile to add nodejs

- Adds Tailwind CSS settings to the `Web` app

- Adds tailwind tags to the base template

- Uses tailwind classes in the homepage template


docker/django/Dockerfile
```diff
@@ -91,14 +95,18 @@ CMD python manage.py rqworker --with-scheduler
 FROM python-base as web-dev
 
 CMD python /opt/bigcases2/manage.py migrate  && \
+    python /opt/bigcases2/manage.py tailwind install --no-input  && \
```
mlissner, line 98: Why do these need to get installed every time? Aren't they part of our code base?

mlissner (changes requested):
A couple things on this one:

1. What about `django_browser_reload`? 

3. What about the `tailwind_preload_css` template tag? 

Looks pretty good. A couple comments, then I want to give it a try myself. Which command actually runs npm?

docker/django/Dockerfile
ERosendo: This is the command that runs the `npm install` and creates a `node_modules` folder with all the dependencies. We're ignoring this folder so We need to create it before We use the `build` command

ERosendo:
Thanks for your comments. 

1. I just added `django_browser_reload`, so the browser will automatically reload when static assets or templates are modified. 

2. I also added the `tailwind_preload_css` tag. I read in the docs that improve loading performance in production but I forgot to place it above the `tailwind_css` tag.

I also added the favicon icons following this [guide](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs).

ERosendo:
@mlissner I added a service that runs the Tailwind development server to watch for file changes and rebuild the assets. We won't have to run the `start` command all the time.

I used [this](https://github.com/timonweb/django-tailwind/blob/master/example/docker-compose.yml) example to update the `docker-compose.yml` file.

mlissner (approved):
LGTM.

Issue #110, opened 2023-03-02
Automatically compile and deploy tailwind


mlissner:
Over in #109, we tweaked things to manually add style.css and to expect the developer to remember to do that. I went ahead and merged it so we can get the site up and running again, but it seems to me like the wrong solution. 

In CourtListener, where we use react, developers are expected to generate minified javascript and commit it to the repo. It kinda stinks and is easy to forget. We need less of that kind of thing and I'm not happy to see a similar kind of "Don't forget to" gotcha here, where it's a greenfield that can be done right.

@ERosendo, do you think instead of the manual process you could make this part of the Github Action? I guess you'd just need to install NPM and then run the build command, before running collectstatic?

PR #118, opened 2023-03-07
feat(github action): Compile tailwind styles before collecting static files.

ERosendo:
This PR adds the commands to compile the tailwind style to the GitHub action that collects static assets. This PR fixes https://github.com/freelawproject/bigcases2/issues/110.


mlissner:
Shouldn't this remove the CSS file that's committed already?

ERosendo:
You're right! I will remove the CSS file We added in https://github.com/freelawproject/bigcases2/pull/109

Issue #32, opened 2023-01-27
Add Plausible to the Web site


mlissner:
This should be easy too. We just need to generate a new web token for it, and put that in our settings file, like with CL.

mlissner:
Here's the snippet Plausible gives us for the site (but it's worth comparing it against the one in CL to see if there are any useful customizations vs the default): 

```html
<script defer data-domain="bots.law" src="https://plausible.io/js/script.js"></script>
```

PR #123, opened 2023-03-08
feat(web): Add Plausible to the website

ERosendo:
This PR adds the Plausible's snippet to the `base` template of the project. This PR fixes https://github.com/freelawproject/bigcases2/issues/32


ERosendo:
Sure!

ERosendo:
@mlissner I've removed the tag that creates the random int

bc/assets/templates/base.html
```diff
@@ -1,4 +1,4 @@
-{% load static tailwind_tags %}
+{% load static %}{% load tailwind_tags %}
```
mlissner, line 1: I'm pretty sure you can do this in newer versions of Django:

```suggestion
{% load static tailwind_tags %}
```

mlissner (changes requested):
One suggestion, please. Go ahead and merge once you've looked into it, and we can verify that it's working by visiting plausible.com.

bc/assets/templates/base.html
ERosendo: You're right. We can do that with the `load` tag

mlissner (approved):

PR #126, opened 2023-03-09
fix(web): Tweak homepage accessibility

ERosendo:
This PR fixes accessibility issues from PageSpeed Insights.

- Adds [alt] attributes to the image elements

- Use darker text in the footer

- Use the fallback font to display the text during webfont load

### DIFF
--- a/bc/web/templates/homepage.html
+++ b/bc/web/templates/homepage.html
@@ -9,7 +9,7 @@
             <h1 class="md:max-w-md xl:max-w-xl font-[800] text-5xl md:text-6xl lg:text-6xl xl:text-7xl text-left">
                 Hi! I’m the bot that keeps you updated about court cases!
             </h1>
-            <img src="{% static "svg/full-body-left-hand.svg" %}" class="hidden sm:block w-32 sm:w-56 md:w-80 lg:w-96 xl:w-[500px]">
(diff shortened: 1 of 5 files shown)


bc/web/templates/homepage.html
```diff
@@ -20,7 +20,7 @@ <h1 class="md:max-w-md xl:max-w-xl font-[800] text-5xl md:text-6xl lg:text-6xl x
         <div class="py-8 md:py-16 text-white">
             <div class="grid grid-cols-2 flex flex-col justify-center gap-10 pt-6">
                 <div class="flex justify-start self-start">
-                    <img src="{% static "svg/chat-app-logos-combined.svg" %}" class="w-48 object-contain sm:w-64 md:w-72 lg:w-80 xl:w-[490px] xl:h-[440px]">
+                    <img alt='Chat app logos combined' src="{% static "svg/chat-app-logos-combined.svg" %}" class="w-48 object-contain sm:w-64 md:w-72 lg:w-80 xl:w-[490px] xl:h-[440px]">
```
mlissner, line 23: We can do better here. How about "logos for popular chat applications, including Slack, Discord, Microsoft Teams and Google Chat"

mlissner (changes requested):
A couple tweaks, but looks good to me otherwise.

bc/web/templates/homepage.html
ERosendo: You're right! That's a lot better

mlissner (approved):

PR #132, opened 2023-03-10
feat(masto): Tweak post template

mlissner:
This tweaks the mastodon template a bit to make it a bit clearer. There are a lot of cases that are named "In re...". When we got those, we'd tweet something like, `New filing in In re...`, which is confusing. 

This tweaks it to be `New filing: "In re...."`, which should be better and I think is no longer.


ERosendo (approved):
LGTM

Issue #119, opened 2023-03-07
Make the new website good


mlissner:
Over in #113, we're doing some clean up stuff, but we still have a lot to do to make the site really good. Among them:

 - Compile and deploy tailwind in our CI pipeline (#110)
 - Invalidate cache on deployment (#114)
 - Make the PageSpeed insights happy (#116)

Once that's done, it's onwards to all the pages the site needs. Most of these will just be text with a picture or two, but some will need dynamic content (these will need to avoid being cached by CloudFront via a django decorator on the view). 

Anyhow, these are the pages and components we need:

  - [x] The Big Cases header drop down should offer two drop down options: 
      - "About" (`/big-cases/about/`): This will have information about the history of the bot, the curators, and the list of cases followed.
      - "Sponsors" (`/big-cases/sponsors/`): This will have information about the Sponsorship program, past sponsors, etc.

    We'll need to update the footer to just link to these two places as well.

 - [x] The Little Cases header will just go to `/little-cases/`, which will have a simple page explaining that we aim to let other people set up bots too, and providing a form for people to "Suggest a Bot". The form could just ask the person's name, preferred platforms (checkboxes for mastodon, or twitter) and a free text box that asks them for the kinds of cases it would follow.

    Eventually, when we have our first Little Cases bot, we'll have to provide links saying who maintains each, what cases each is following, etc. 

    At first, it'll probably be hands on creating Little Cases bots, but ideally we'll make it totally hands off over time.
 
 - [x] The Collaboration Suites header goes to `/collaboration/` and provides a page explaining the features we aim to have in Slack, Discord, etc. This is something of a product page, and it has a form to sign up for the waitlist. The form asks for your preferred platforms (slack, discord, MS Teams, or Google Chat), contact info (name and email), and company name (optional)

 - [x] Donate link is good to go for now.

 - [x] Sign In link should go away for now. We're not ready for it, but we should at least get `/sign-in/` working so we can get the django admin page up and running. We'll eventually need the full suite of user-handling junk, unfortunately, but we can put it off for now since it involves setting up email and everything else. We'll need it for Discord/Slack/etc though.

 - [x] The Join the Waitlist button simply takes you to `/collaboration/#join`, and we can put a node with `id="join"` on the form.

 - [x] The Learn More links:
     - The one under Big Cases takes you to /big-cases/about/
     - The one under Little Cases takes you to /little-cases/
     - The one under Free, Searchable Documents takes you to `/big-cases/my-code/`

[...]

PR #138, opened 2023-03-15
feat(templates): Remove links and sections from the footer

ERosendo:
This PR tweaks the footer template following the suggestions from https://github.com/freelawproject/bigcases2/issues/119 and also updates the styles to be consistent with the design.


mlissner:
Looking good!

Issue #95, opened 2023-02-27
Add new column for the case summary/gloss/blurb


mlissner:
A problem I'm having is that case names often aren't clear what they're about. For example, "Sweet v. Cardona" is the student loan debt relief case, but the case name doesn't really tell you that. So when you see:

> New filing in Sweet v. Cardona ...

Who knows what that's talking about.

I've been handling this a little bit by putting the a short blurb into the case name field itself, so for that one I might name the case:

> Sweet v. Cardona (student debt relief)

That's messy. Instead, let's add a column to the DB for this and prompt for it (optionally) in the `lookup --add` command. Then in our posts, we can put it in parentheses, as above, after the case name, if it exists.

We probably want to put a max length on this one. Maybe 30 chars?

Let's make sure we do this before we do #24 too, so that when we let others start maintaining this, it's already in place.

ERosendo:
You mentioned something like this when We reviewed https://github.com/freelawproject/bigcases2/issues/20 and We added a field called `case_summary` in the `Subscription` model. Here's the code:

https://github.com/freelawproject/bigcases2/blob/4a20113150731cb724177e1ba62485c900b9563b/bc/subscription/models.py#L37-L41

We're not using this field in the current implementation of the `lookup` command and I think the `help_text` matches the description of the new column. 

We can change the `max_length` of this field, tweak the `lookup` command to prompt for it and start saving data in this field.

let me know what you think.

mlissner:
Ah! I thought we cut out that column. Thanks for the reminder. So let's just start using that in lookup and in our posts. No need to tweak the max_length. It's good enough as is. 

But let's do this later. I'll get it prioritized in your backlog.

PR #140, opened 2023-03-16
feat(subscription): Prompt for case summary in the lookup command

ERosendo:
This PR fixes https://github.com/freelawproject/bigcases2/issues/95

- Prompt for a case summary in the `lookup --add`

- Adds a new property to the `subscription` model that adds the **summary**(when its available) to the **docket name**.

- Use the new property in the mastodon templates.


mlissner (approved):
LGTM!

mlissner:
Once we get the admin panel up and running, we need to update a few cases that have the summary in the docket name, but I think we'll be OK not doing so immediately.

PR #141, opened 2023-03-17
feat(web): Remove Sign In link and set up admin login

ERosendo:
This PR removes the Sign In link on the header, adds django-ratelimit as a dependency, and redirects the admin login to a custom login page that uses a ratelimiting decorator. 

Here's a screenshot of the login page:

image

### DIFF
--- a/bc/core/utils/network.py
+++ b/bc/core/utils/network.py
@@ -1,7 +1,8 @@
-from typing import Optional
+from django.http import HttpRequest
+from django_ratelimit.core import get_header
 
 
-def strip_port_to_make_ip_key(ip_str: str | None) -> str | None:
+def strip_port_to_make_ip_key(group: str, request: HttpRequest) -> str | None:
     """Make a good key to use for caching the request's IP
 
     CloudFront provides a header that returns the user's IP and port,
@@ -18,6 +19,5 @@ def strip_port_to_make_ip_key(ip_str: str | None) -> str | None:
     :param ip_str: the IP address of the viewer and the source port of the request
     :return: A simple key that can be used to throttle the user if needed.
     """
-    if ip_str:
-        return ip_str.split(":")[0]
-    return None
+    header = get_header(request, "CloudFront-Viewer-Address")
+    return header.split(":")[0]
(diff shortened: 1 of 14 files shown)


bc/core/utils/network.py
```diff
@@ -18,6 +20,13 @@ def strip_port_to_make_ip_key(ip_str: str | None) -> str | None:
     :param ip_str: the IP address of the viewer and the source port of the request
```
mlissner, line 20: The docstring needs a few tweaks here.

bc/users/templates/register/login.html
```diff
@@ -0,0 +1,33 @@
...
+    <h1 class="font-display font-light tracking-wide">Sign In</h2>
+    {% if form.errors %}
+        {%  for error in form.non_field_errors %}
+        <p class="alert alert-danger">{{ error|safe }}</p>
+        {% endfor %}
+    {% endif %}
+
+    <form action="{% url "sign-in" %}" method="post">
+        {% csrf_token %}
+    <div class="grid grid-cols-1 gap-4">
```
mlissner, line 17: Indentation is a bit weird on this div. Also, we should be doing two spaces, not four in HTML, according to our editorconfig.

bc/users/urls.py
```diff
@@ -0,0 +1,20 @@
...
+
+urlpatterns = [
+    # Sign in page
+    path(
+        "sign-in/",
+        ratelimiter_unsafe_10_per_m(
+            auth_views.LoginView.as_view(
+                **{
+                    "template_name": "register/login.html",
+                    "extra_context": {"private": False},
```
mlissner, line 14: We're not using this yet in our base template, are we? If not, we should either remove it from here or add it to the base template. I think we can probably just remove it here and we won't have private pages in this project?

mlissner (changes requested):
Good stuff. Looks good too. A few comments, thank you!

bc/users/urls.py
ERosendo: No, We're not using the "private" attribute in the base.html. I'll remove it.

bc/core/utils/network.py
ERosendo: Thanks!. I'm going to update the docstring.

mlissner (approved):

Issue #23, opened 2023-01-26
Add thumbnails for documents that are available when we post


mlissner:
I'm not sure if this is already ready, but we can launch without it, so I'm setting it as a post launch milestone that we can do later.

mlissner:
I think the trick for this one is to build in a small delay between when we get a webhook event from CL and when we send our social posts. During that delay, somebody might buy the document and if they do, we can download it, make thumbnails, etc. If the delay is configurable, we can tweak it to see how big of a delay actually helps.

ERosendo:
We're using the `enqueue_in` method to [schedule](https://github.com/rq/rq-scheduler#scheduling-a-job) the execution of the function that sends the social posts. 

We also have a variable in the .env file called `WEBHOOK_DELAY_TIME` to set the delay in seconds. Here's the code We're using to schedule a task:

https://github.com/freelawproject/bigcases2/blob/bcbf9f64026882637e685303a6913bb376b06a75/bc/subscription/api_views.py#L59-L63

Issue #17, opened 2022-12-09
Add captions/alt-text to document thumbnails


anseljh:
Captioning is great, and also a stronger convention in Mastodon. 

See, e.g., https://botsin.space/@PleaseCaption.

mlissner:
We can only do this for documents we have when we send our toot/tweet, but we should build it in early because it's pretty easy. 

We could use doctor to get the text, but I think it's probably more performant to get it from the CL API.

mlissner:
Actually, I don't think we should bother getting the text of the documents. It's too much trouble and people will have the link to the doc right there. We can just say:

> Thumbnail of page 1

> Thumbnail of page 2

Etc.

PR #159, opened 2023-03-28
feat(channels): Add thumbnails for documents

ERosendo:
This PR adds the logic to create thumbnails from a document and attach them to a post. It also adds `alt text` to those thumbnails. This PR fixes https://github.com/freelawproject/bigcases2/issues/17 and also fixes https://github.com/freelawproject/bigcases2/issues/23. 

- Bump [twitterAPI](https://github.com/geduldig/TwitterAPI) from 2.8.1 to 2.8.2. This new version of the library will allow us to attach `alt text` to the media uploads.

- Adds the [Doctor](https://github.com/freelawproject/doctor) endpoint to the settings module and also adds a helper function to get thumbnails of a document.

- Adds a new field to the `WebhookEvent` model. This new field is called `cl_doc_id` and stores the document_id from CourtListener. This PR also introduces a helper function to lookup recap documents that use the `CL API` and the new `cl_doc_id` field.

- Refactors the logic to upload a media file. This PR adds a new method called `upload_media` to the `BaseAPIConnector` interface.

- Updates the `process_filing_webhook_event` and uses the new helper functions to attach thumbnails to the new posts when a document is available in the RECAP archive. 

Here are some examples of posts with thumbnails created with the new `process_filing_webhook_event` method:

- This one shows a typical example( We have a short description): 
         https://twitter.com/flp_tests/status/1640735514792800259

- This one shows how it works when the docket entry has a long description:
         https://twitter.com/flp_tests/status/1640702446128201729


bc/channel/utils/connectors/masto.py
```diff
@@ -29,22 +30,42 @@ def get_api_object(self, version=None) -> ApiWrapper:
...
+
+        if thumbnails:
+            for idx, thumbnail in enumerate(thumbnails):
+                media_id = self.upload_media(
+                    thumbnail, f"Thumbnail of page {idx + 1}"
+                )
+                if media_ids:
+                    media_ids.append(media_id)
+                else:
+                    media_ids = [media_id]
```
mlissner, line 68: Why not make medea_ids an empty list up above, and you can simplify this?

bc/channel/utils/connectors/twitter.py
```diff
@@ -48,12 +72,19 @@ def add_status(
...
+                f"An image of the entry's full text: {text_image.description}",
             )
-            media_id = media_response.json()["media_id"]
             media_array.append(str(media_id))
+            payload["media"] = {"media_ids": media_array}
 
+        if thumbnails:
+            for idx, thumbnail in enumerate(thumbnails):
+                media_id = self.upload_media(
+                    thumbnail, f"Thumbnail of page {idx + 1}"
```
mlissner, line 85: Here and above, it'd be a bit better to say:

```suggestion
                    thumbnail, f"Thumbnail of page {idx + 1} of the PDF"
```

bc/core/utils/microservices.py
```diff
@@ -0,0 +1,33 @@
...
+        page_range (str): str representation of the list of pages requested
+
+    Returns:
+        list[bytes]: list of thumbnails
+    """
+    thumbnails = requests.post(
+        f"{settings.DOCTOR_HOST}/convert/pdf/thumbnails/",
+        data={"pages": page_range, "max_dimension": "1920"},
+        files={"file": (f"dummy.pdf", document)},
+        timeout=4 * 60,
```
mlissner, line 27: This is very long to do some simple PDF work. I would expect this to complete in less than a minute? If we have retries anyway, it's probably better to make this a minute and if that's not enough, let the retry take care of it.

bc/subscription/models.py
```diff
@@ -125,6 +125,10 @@ class FilingWebhookEvent(AbstractDateTimeModel):
         help_text="The docket id from CL.",
         null=True,
     )
+    cl_doc_id = models.IntegerField(
+        help_text="The document id from CL.",
+        null=True,
+    )
```
mlissner, line 131: Calling this `cl_doc_id` is inconsistent vs the docket_id just above. Why isnt' that one cl_docket_id, or why isn't this one just doc_id?

bc/subscription/utils/courtlistener.py
```diff
@@ -67,6 +68,20 @@ def lookup_docket_by_cl_id(cl_id: int):
     return response.json()
 
 
+def lookup_document_by_doc_id(cl_doc_id: int | None):
```
mlissner, line 71: When would this be `| None`? What type does this function return?

bc/subscription/tasks.py
```diff
@@ -38,6 +41,19 @@ def process_filing_webhook_event(fwe_pk) -> FilingWebhookEvent:
...
+    cl_document = lookup_document_by_doc_id(filing_webhook_event.cl_doc_id)
+    document_url = (
+        f"https://storage.courtlistener.com/{cl_document['filepath_local']}"
+        if cl_document["filepath_local"]
+        else None
+    )
+
+    document = None
+    if document_url:
+        document_request = requests.get(document_url, timeout=60)
```
mlissner, line 53: You're fetching this from S3. A timeout of maybe 3 seconds would certainly be fine.

mlissner (dismissed):
A few little things, but generally looks good. I feel like it's getting to the point where I can no longer fit it all in my head. It was fun while it lasted!

bc/subscription/tasks.py
ERosendo: got it!

bc/core/utils/microservices.py
ERosendo: understood!. I will tweak the timeout.

bc/channel/utils/connectors/masto.py
ERosendo: I thought the API could throw an error if I used an empty list but I tried your suggestion and It works fine.

bc/channel/utils/connectors/twitter.py
ERosendo: got it! I will update the `alt text` for the thumbnails

bc/subscription/models.py
ERosendo: You're right, We're not using a namespace for the fields in this model. We should call it `doc_id`.

bc/subscription/utils/courtlistener.py
ERosendo: The `doc_id` won't be `None` for new records but I set the default value to `null` for old records (We need a default value because this table has data). 

Mypy was complaining about the type hint because the `doc_id` field is nullable in the `WebhookEvent` model and that's why I included `None` in the type hint of the helper function. 

This helper function returns a `dict` and I realized We can use the `fields` parameter to include just the `filepath_local` attribute in the response. The new request would be:

```python
response = requests.get(
        f"{CL_API['recap-documents']}{doc_id}/",
        params={"fields": "filepath_local"},
        headers=auth_header(),
        timeout=5,
    )
```

Using the previous request, the return type would be: `dict[str, str | None]`

Do you think We can use another approach to set a default value for previous records?

PR #165, opened 2023-03-30
feat(sponsorship): Add app to track sponsorship values. 

ERosendo:
This PR adds a new app to the project called `sponsorship`. 

This new app introduces two models:
 - `Sponsorship`: stores new sponsorships and keeps track of the amount of money left to buy new documents.
 - `Transactions`: stores all the transactions related to sponsorships ( new sponsorships, document purchases, and adjustments) 

This PR register both models to the admin site and also introduces a receiver function to handle the post_save signal of the Sponsorship model.


bc/sponsorship/models.py
```diff
@@ -0,0 +1,79 @@
...
+        (SPONSORSHIP, "Sponsorship"),
+        (DOCUMENT_PURCHASE, "Document Purchase"),
+        (ADJUSTMENT, "Adjustment"),
+    )
+
+    user = models.ForeignKey(
+        User,
+        help_text="User associated with the transaction",
+        related_name="transactions",
+        on_delete=models.CASCADE,
```
mlissner, line 47: We shouldn't delete transactions. Instead of CASCADE, how about PROTECT? 

That should prevent the user from being deleted if they have transaction records. In that case, we'd want to keep the user around, but set the user as a stub account or similar.

bc/sponsorship/models.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    sponsorship = models.ForeignKey(
+        Sponsorship,
+        help_text="Sponsorship record related to the transaction",
+        related_name="transactions",
+        on_delete=models.CASCADE,
```
mlissner, line 53: Same applies here. Deleting a sponsorship that has been spent should not be allowed.

bc/sponsorship/models.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.pk}: {self.user.username} - {self.original_amount}"
+
+
+class Transaction(AbstractDateTimeModel):
+    SPONSORSHIP = 1
+    DOCUMENT_PURCHASE = 2
+    ADJUSTMENT = 3
+    CHOICES = (
```
mlissner, line 37: CHOICES is a weird variable name. If you wind up using it in other code, you'd have:

```
Transaction.CHOICES
```

Not great. Perhaps:

```
Transactions.TYPES
```

That's dangerously close to a reserved word in Python, but frankly, I worry about your use of `type` as the field name too, but I think it should be fine.

bc/sponsorship/models.py
```diff
@@ -0,0 +1,79 @@
...
+    def __str__(self) -> str:
+        match self.type:
+            case self.SPONSORSHIP:
+                return f"{self.pk}: New sponsorship"
+            case self.DOCUMENT_PURCHASE:
+                return f"{self.pk}: New document purchased"
+            case self.ADJUSTMENT:
+                return f"{self.pk}: Adjustment"
+            case _:
+                return f"{self.pk}"
```
mlissner, line 79: Less is more. How about:

```suggestion
        return f"{self.pk}: {self.get_type_display()}"
```

bc/sponsorship/selectors.py
```diff
@@ -0,0 +1,5 @@
+from .models import Sponsorship
+
+
+def get_available_sponsorship() -> Sponsorship | None:
```
mlissner, line 4:
```suggestion
def get_active_sponsorship() -> Sponsorship | None:
```

bc/sponsorship/selectors.py
```diff
@@ -0,0 +1,5 @@
...
+    return Sponsorship.objects.filter(current_amount__gte=3.00).first()
```
mlissner, line 5: This has magic happening at a distance by relying on the `Meta.order` property. I think it'd be a lot better to put that explicitly here and to remove it from the model. I tend to see default ordering as an anti-pattern because you might rely on it, but you might not, maybe it's doing something, maybe not...you can never be sure and thus you can never change it.

bc/sponsorship/models.py
```diff
@@ -0,0 +1,79 @@
...
+        (ADJUSTMENT, "Adjustment"),
+    )
+
+    user = models.ForeignKey(
+        User,
+        help_text="User associated with the transaction",
+        related_name="transactions",
+        on_delete=models.CASCADE,
+    )
+    sponsorship = models.ForeignKey(
```
mlissner, line 49: This should be nullable, right? For adjustments that may not relate to a sponsorship?

mlissner (changes requested):
Cool. I have lots of thoughts. Let me know what you think.

bc/sponsorship/models.py
ERosendo: understood!

bc/sponsorship/models.py
ERosendo: Thanks for the suggestion. I'll use `PROTECT` instead.

bc/sponsorship/models.py
ERosendo: got it! Thanks for your suggestion.

bc/sponsorship/models.py
ERosendo: Awesome! Thanks for your suggestion.

bc/sponsorship/models.py
ERosendo: You're right!

bc/sponsorship/selectors.py
ERosendo: got it! I'll use the order_by() method instead.

ERosendo:
Thanks for your comments. I've applied your suggestions.

mlissner:
I appreciate how you did just the models in this PR and the logic comes later. That made it a lot easier to think about.

Issue #16, opened 2022-12-06
Automatically purchase documents from PACER for cases we follow


anseljh:
https://www.courtlistener.com/api/rest-info/#recap-endpoint

See #7

mlissner:
We will eventually want to start purchasing documents using a budget set by Free Law Project. When that time comes, we'll do it using the document fetch API from CourtListener.

PR #179, opened 2023-04-04
feat(subscription): Purchase documents from PACER

ERosendo:
This PR adds logic to automatically purchase documents from PACER and fixes https://github.com/freelawproject/bigcases2/issues/16.

- Add the page_count attribute to the lookup_document request

    This PR tweaks the helper method called `lookup_document_by_doc_id` to add the `page_count` attribute in the response of the CL API and also updates the type hinting of the helper method. We need the `page_count` of the document to compute its price. 

- Adds a helper method to purchase documents using the CL API.
  
    This helper uses the `recap-fetch/` endpoint and the PACER's login credentials to purchase PDFs.

- Adds a helper method to download documents from the RECAP archive.

- Tweaks the helper method called `process_filing_webhook_event` to add the logic to purchase PDF documents if they're not available in the RECAP archive.

- Add a task to handle RECAP fetch webhooks

- Adds a method to handle the transaction's post_save signal. We're using this signal to update the `current_amount` in the `sponsorship` model.

I know the current implementation of the bot is a little hard to fit all in one’s head but the new implementation of the task that handles the docket alert webhooks will work like this (after we add sponsorships and document purchases): 

- The bot gets a docket alert webhook and the main task starts.

- The main task links the webhook to a subscription, checks the webhook description, and tries to retrieve the document. We can have a few scenarios here:

     - If the document is available, The bot downloads it, creates the thumbnails and the main task enters the loop and starts scheduling the subtasks to create the posts.
     
     - If the document is not available but there are sponsorships available, the main task uses the `recap-fetch `endpoint to purchase the PDF and finishes its execution ( The bot will use the RECAP fetch webhook to create the posts)
    
     - if the document is not available and there's no sponsorship available, the main task enters the loop and starts scheduling the subtasks to create the post without thumbnails.

### DIFF
--- a/bc/sponsorship/signals.py
+++ b/bc/sponsorship/signals.py
@@ -1,7 +1,9 @@
+from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
 from .models import Sponsorship, Transaction
+from .utils import update_sponsorships_current_amount
 
 
 @receiver(post_save, sender=Sponsorship)
@@ -14,4 +16,16 @@ def sponsorship_handler(sender, instance=None, created=False, **kwargs):
             user=instance.user,
             sponsorship=instance,
             amount=instance.original_amount,
+        )
+
+
+@receiver(post_save, sender=Transaction)
+def transaction_handler(sender, instance=None, created=False, **kwargs):
+    if created and instance.type == Transaction.DOCUMENT_PURCHASE:
+        """
+        make sure We update the sponsorship's current_amount only if
+        the django atomic transaction successfully commits.
+        """
+        transaction.on_commit(
+            lambda: update_sponsorships_current_amount(instance)
         )
--- a/bc/sponsorship/signals.py
+++ b/bc/sponsorship/signals.py
@@ -1,4 +1,4 @@
-from django.db import transaction
+from django.db import transaction as db_transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
@@ -26,6 +26,6 @@ def transaction_handler(sender, instance=None, created=False, **kwargs):
         make sure We update the sponsorship's current_amount only if
         the django atomic transaction successfully commits.
         """
-        transaction.on_commit(
+        db_transaction.on_commit(
             lambda: update_sponsorships_current_amount(instance)
         )
--- a/bc/sponsorship/utils.py
+++ b/bc/sponsorship/utils.py
@@ -1,11 +1,22 @@
 from decimal import Decimal
 
-from .models import Sponsorship, Transaction
+from django.db.models import F
+
+from .models import Transaction
 
 
 def update_sponsorships_current_amount(ledger_entry: Transaction) -> None:
-    if not ledger_entry.sponsorship_id:
+    """
+    This method updates the current_amount field of the sponsorship record
+    related to the given ledger entry after a document has been purchased.
+
+    Args:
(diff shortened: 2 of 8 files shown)


bc/sponsorship/signals.py
```diff
@@ -1,7 +1,9 @@
+from django.db import transaction
```
mlissner, line 1: This is confusing. Maybe do something like:

```suggestion
from django.db import transaction as db_transaction
```

bc/subscription/tasks.py
```diff
@@ -53,18 +58,57 @@ def process_filing_webhook_event(fwe_pk: int) -> FilingWebhookEvent:
...
-        else None
+    if cl_document["filepath_local"]:
+        document = get_document_from_CL(cl_document["filepath_local"])
+    else:
+        sponsorship = get_active_sponsorship()
+        if sponsorship and filing_webhook_event.pacer_doc_id:
+            purchase_pdf_by_doc_id(filing_webhook_event.doc_id)
+            return filing_webhook_event
+
+    for channel in get_enabled_channels():
```
mlissner, line 71:
```suggestion
    # Got the document or no sponsorship. Tweet and toot.
    for channel in get_enabled_channels():
```

bc/subscription/tasks.py
```diff
@@ -53,18 +58,57 @@ def process_filing_webhook_event(fwe_pk: int) -> FilingWebhookEvent:
...
+    sponsorship = get_active_sponsorship()
+    if sponsorship:
+        Transaction.objects.create(
+            user=sponsorship.user,
+            sponsorship=sponsorship,
+            type=Transaction.DOCUMENT_PURCHASE,
+            amount=3.0
+            if cl_document["page_count"] >= 30
+            else cl_document["page_count"] * 0.10,
+        )
```
mlissner, line 111: This feels like it might be well placed into a function. Something like:

```
def log_purchase(...)
```

Also, we have a note field. Let's put something in the note explaining what was bought with docket/document/court info if we have it?

mlissner (changes requested):
Looks good. A few things.

bc/sponsorship/utils.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .models import Sponsorship, Transaction
+
+
+def update_sponsorships_current_amount(ledger_entry: Transaction) -> None:
+    if not ledger_entry.sponsorship_id:
+        return None
+    sponsorship = Sponsorship.objects.get(pk=ledger_entry.sponsorship_id)
+    sponsorship.current_amount -= Decimal(ledger_entry.amount)
+    sponsorship.save()
```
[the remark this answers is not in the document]
ERosendo: Thanks for your comments and sorry about this code 😱

1. Yes, We can use the relationship to get the sponsorship.
2.  Sure. I will use an F expression to update the `current_amount`.
3. Alright! I will add a docstring.

bc/sponsorship/signals.py
ERosendo: understood!

bc/subscription/tasks.py
ERosendo: You're right about the logic to create a new entry in the transaction table, I will refactor that code into a helper.

ERosendo:
Thanks for your comments. I've applied your suggestions.

Let me know what you think.

mlissner (approved):

PR #181, opened 2023-04-06
feat(sponsorship): Add text to sponsored documents thumbnails.

### LINKED ISSUE #151: Add watermarks and thumbnails to sponsored documents
When we start buying documents in #16, we need to add watermarks to them. It'll work like this:

1. If it's sponsored by an organization, the documents will get a watermark on them that says the organization's name: "Sponsored by ".

2. If it's sponsored by FLP, there's no watermark, but any document with fewer than 4 pages gets an extra thumbnail at the end. That thumbnail says that the document is sponsored by Free Law Project, provides a link to the sponsorship page (maybe a QR code?) and encourages others to sponsor the bot.
    - If the document has four or more pages, we don't do this, and we provide the regular thumbnails with no watermark.

3. If the document isn't sponsored at all, then we can do a thumbnail every so often saying that we need sponsors, but I think it's more elegant to handle this by email, and we should do it that way. So in this case, no thumbnail at all, just like we do now.

I think that should do the trick.

### LINKED ISSUE #168: Add `message` (?) field to the sponsorship table
We just landed the sponsorship table in #152, but we forgot an important column. The current system would allow us to say that:

 - User 22 sponsored this document

Or:

 - User RECAP sponsored this document

That's nice, but wouldn't it be better if there were a better message, like:

 - This case sponsored by XYZ Law

Or:

 - Eat spinach every day to keep the shave away — [Burma Shave](https://en.wikipedia.org/wiki/Burma-Shave)

Let's add a field to the database where we can store that message. Perhaps in the sponsorship table? Perhaps the field can be named `message` or `watermark_message`?

ERosendo:
This PR adds the logic to include custom messages in thumbnails of sponsored .documents This PR fixes #151  and fixes #168. 

- Adds a new field called `watermark_message` to the sponsorship model. 

     - This new field allows us to show custom messages in thumbnails of sponsored documents. This PR also adds a new property called `thumbnail_message`, which returns a default message if the sponsor doesn't provide a custom text to show in the thumbnails. 

- Adds a new class to handle thumbnails of sponsored documents.

     - This new class has a few helpers methods that allow us to draw text along the left edge of a thumbnail and get the final result as a bytes object.

- Adds a new helper method to add the sponsor message to the thumbnails. 

     - This helper method receives the list of thumbnails and a string and creates instances of the `SponsoredThumbnail` class.

- Updates the **return type** of the method called `get_thumbnails_from_range`, so **Mypy** won't complain about the thumbnail data type when the bot adds the sponsor's message. 
 
     - The current implementation of the method returns the array of thumbnails that we get from [Doctor](https://github.com/freelawproject/doctor) as a list of `IO objects` and We use that list when the document is available in the RECAP archive(the type hint in the API classes is `list[IO[bytes]]`), but We end up with a different list(`list[bytes]`) when the bot purchases the document and adds a message because the `SponsoredThumbnail` class creates bytes objects instead of IO[bytes] objects, so the logic in the loop that unzip the thumbnails in the `get_thumbnails_from_range method` is updated to create a  `list[bytes]` instead of a `list[IO[bytes]]`.

- Registers the `User` model in the admin interface.

- Alter the affiliation field in the User models. 

    - I tried to update one of the records in the user table using the admin panel( I was trying to update the `first_name` and the `last_name` fields) and noticed this field was required, so This PR sets the **blank** attribute to **True** to allow empty values for this field. 
  
Here are some tests that show the thumbnails for sponsored documents:

- Thumbnails with the default message: https://twitter.com/flp_tests/status/1644407917305634818

- Thumbnail with a custom message: https://twitter.com/flp_tests/status/1644409981234552850


bc/sponsorship/models.py
```diff
@@ -22,10 +22,21 @@ class Sponsorship(AbstractDateTimeModel):
...
+    )
 
     def __str__(self) -> str:
         return f"{self.pk}: {self.user.username} - {self.original_amount}"
 
+    @property
+    def thumbnail_message(self) -> str:
+        if self.watermark_message:
+            return self.watermark_message
+        return f"This document contributed by {self.user.name}"
```
mlissner, line 38: Hm, I think it's worth just requiring the `watermark_message` field as part of every sponsorship. I think sometimes we might want self.user.name, other times, we'd want `self.user.affiliation`, or something else. It's probably just worth thinking about it and not having a default that might be wrong.

bc/subscription/tasks.py
```diff
@@ -101,7 +102,9 @@ def process_fetch_webhook_event(fwe_pk: int):
     document = download_pdf_from_cl(cl_document["filepath_local"])
 
     sponsorship = get_active_sponsorship()
+    sponsor_message = None
```
mlissner, line 105: Something to think about, and maybe fix eventually: This function is getting pretty "thick." In theory, this function should be very "thin," leaving just about all logic and look-ups to the task, so it can respond to the webhook event as quickly and simply as possible.

I think it'd be pretty easy to move most of this logic into the task, and make the task's signature much smaller. It's usually code-smell when you have five required parameters to a function.

We can fix it later, but it's something I noticed.

bc/subscription/tasks.py
```diff
@@ -124,7 +128,11 @@ def process_fetch_webhook_event(fwe_pk: int):
 
 @transaction.atomic
 def make_post_for_webhook_event(
-    channel_pk: int, subscription_pk: int, fwe_pk: int, document: bytes | None
+    channel_pk: int,
+    subscription_pk: int,
+    fwe_pk: int,
+    document: bytes | None,
+    sponsor_text: str | None = None,
```
mlissner, line 135: Needs docstring.

bc/users/migrations/0002_alter_user_affiliation.py
mlissner, line 1: This is a noop on the database, so I'd add `_noop.py` to the name, but I know that's a hassle now. Something to remember in the future as a good convention.

bc/core/utils/images.py
```diff
@@ -278,3 +278,163 @@ def to_bytes(self) -> bytes:
...
+        the sponsored text.
+
+        Returns:
+            tuple[int,int]: (width, height) dimensions
+        """
+        title_w, title_h = self.title_font.getbbox(self.title)[-2:]
+        small_w, small_h = self.small_font.getbbox(self.small_text)[-2:]
+
+        bbox_height = title_h + small_h
+        bbox_width = title_w if title_w >= small_w else small_w
```
mlissner, line 321:
```suggestion
        bbox_width = max(title_w, small_w)
```

mlissner (changes requested):
A few comments, but I think this is basically there! Just gotta get it approved by the board!

bc/core/utils/images.py
ERosendo: You're right. I'll use this built-in function instead.

bc/subscription/tasks.py
ERosendo: you're right. I'll include this parameter in the docstring.

bc/users/migrations/0002_alter_user_affiliation.py
ERosendo: Thanks for your suggestion. I will rename this file.

bc/sponsorship/models.py
ERosendo: understood!

bc/subscription/tasks.py
ERosendo: I agree with you about the number of parameters. I'm not sure if you created the comment in the right function because this one requires just one parameter, I think you're referring to the function called `make_post_for_webhook_event`.

I also noticed you mentioned the webhook event in your comment but none of the functions in this file handle the webhook directly, all of them are used as functions that We invoke asynchronously in a worker (background) process after We store the data of the event.  You can find the functions that receive the webhook events in the [api_views.py file](https://github.com/freelawproject/bigcases2/blob/main/bc/subscription/api_views.py) inside the [subscription folder](https://github.com/freelawproject/bigcases2/tree/main/bc/subscription)

mlissner: > I think you're referring to the function called make_post_for_webhook_event.

Yeah, sorry, it wouldn't let me highlight unedited lines. 

> I also noticed you mentioned the webhook event in your comment

Oh good point. I'll take another look next time we're in this area of the code.

bc/sponsorship/models.py
mlissner: Sorry, to be more explicit, I think we shouldn't allow blank in the model.

ERosendo: got it! Sorry about this. I've removed the `blank` attribute and updated the migration file

ERosendo:
@mlissner I've applied your suggestions. Thanks for your comments.

mlissner (approved):

PR #183, opened 2023-04-06
feat(templates): Add explicit width and height to img elements

ERosendo:
This PR adds the `height` and `width` properties to the image elements in the HTML templates. These properties fix the following issue from the [PageSpeed report](https://pagespeed.web.dev/analysis/https-bots-law/gzo5arnl2a):

- Image elements do not have explicit width and height.


mlissner:
Huh. I thought you'd need to add the height/width attributes to the img tags. I guess we'll see if CSS is enough though.

ERosendo:
I've checked the scores using the "Lighthouse" tab from your google chrome dev tools and the report did not include the issue after using the CSS classes.

mlissner:
I still see this? 

image

mlissner:
Never mind. It's better now. Might have been CloudFront being slow to evict its cache? Who knows. Anyway, we're all good here.

PR #185, opened 2023-04-08
feat(web): Add a rate limit to the pages with forms

### LINKED ISSUE #180: Add captchas (or some other solution) to the forms on the website
I just deleted 300 items from the bot suggestions admin panel. Lots of attempts at SQL injection and the like. Boring stuff we're utterly immune to, but it'd be nice not to have it polluting our admin panel. 

We use HCaptcha on CourtListener, and it works pretty well. we could use it here too, but even just doing a simple test like:

"What's the handle for the Big Cases bot on Twitter?" and making sure people put in `@big_cases` could work. Better, of course, would be to do something funny or clever, but I'm short of that kind of thing right now.

Anyway, we should probably do something about this sooner or later.

ERosendo:
This PR fixes #180 using a rate limit decorator.


ERosendo:
When a request is ratelimited, [django-ratelimit](https://django-ratelimit.readthedocs.io/en/stable/index.html) raises an `exception` that is a subclass of Django’s `PermissionDenied exception`, so the bot serves the text **“403 Forbidden”**, as per [RFC 9110#section-15.5.4](https://datatracker.ietf.org/doc/html/rfc9110.html#section-15.5.4) (the HTTP 1.1 Specification).

We can override the [permission_denied()](https://docs.djangoproject.com/en/4.1/ref/views/#django.views.defaults.permission_denied) view using the [handler403](https://docs.djangoproject.com/en/4.1/ref/urls/#django.conf.urls.handler403) in the root URLconf and catch the ratelimit exception in this custom view to return a different response. The [django-ratelimit](https://django-ratelimit.readthedocs.io/en/stable/index.html) docs show the following example to return a custom response:

```python
def handler403(request, exception=None):
    if isinstance(exception, Ratelimited):
        return HttpResponse('Sorry you are blocked', status=429)
    return HttpResponseForbidden('Forbidden')
```

We can use a custom page that looks like this:

image

Let me know what you think.

mlissner:
Have you checked how CL does this?

ERosendo:
I've just checked how CL is doing this and It's using [HTTP 429 Too Many Requests](https://tools.ietf.org/html/rfc6585#section-4) status code instead of status code 403. The [django-ratelimit](https://django-ratelimit.readthedocs.io/en/stable/index.html) docs also show an example to use this status code.

ERosendo:
@mlissner I've tweaked the `settings` module to use the [HTTP 429 Too Many Requests](https://tools.ietf.org/html/rfc6585#section-4) status code and also included a custom template for this   

I've reviewed the issue with the lint and test action. It seems like the cache is not properly loaded. I've applied the fix described in [this comment](https://github.com/actions/setup-python/issues/182#issuecomment-1151091964) and they're working again.

bc/core/utils/network.py
```diff
@@ -31,3 +31,10 @@ def strip_port_to_make_ip_key(group: str, request: HttpRequest) -> str:
     method=ratelimit.UNSAFE,
     block=True,
 )
+
+ratelimiter_unsafe_1_per_5m = ratelimit(
+    key=strip_port_to_make_ip_key,
+    rate="1/5m",
```
mlissner, line 37: This seems pretty aggressive. Like, I could see myself being annoyed with it, if I had a handful of ideas to input. Maybe 5/30m would be better?

bc/assets/templates/429.html
```diff
@@ -0,0 +1,26 @@
+{% extends "base.html" %}
+
+{% block title %}Forbidden (403){% endblock %}
```
mlissner, line 3:
```suggestion
{% block title %}Sorry, Too Many Requests (429){% endblock %}
```

mlissner (changes requested):
Just a couple little things, thank you!

ERosendo:
@mlissner I've applied the requested changes. Thanks for your comments.

mlissner (approved):
Cool. Looks great, thank you.

PR #192, opened 2023-04-11
fix(core): Check title overflow in the TextImage class

### LINKED ISSUE #191: Title overflows the image
While I was looking for posts related to #190, I found this tweet:

https://twitter.com/big_cases/status/1645463282474680321

In the previous tweet, We can see that the **title** of the image overflows the rectangle. At first, I thought this issue and the one fixed in https://github.com/freelawproject/bigcases2/pull/186 were the same but that's not exactly the case. 

In https://github.com/freelawproject/bigcases2/pull/186, the **description** was the one overflowing the rectangle, and the **title** seems like it could fit the canvas, but this time is the **title** that overflows the canvas and it seems like the **description** could fit if it were centered.  

In the current implementation of the [get_max_character_count](https://github.com/freelawproject/bigcases2/blob/1ad25e01276af345ab588a3d0312957e19ab27cb/bc/core/utils/images.py#L105-L143) method, We're using only the **description** to compute and check the maximum number of characters. This is because the **description** is longer than the **title** and will use most of the space of the image, so I thought that if we found a good number of characters to wrap the **description**, we can use the same number to wrap the **title** too.

The previous approach has worked for most of the images so far(The bot created 22 images yesterday and just one had this issue), but We need to improve it to avoid cases like this.

I think We can fix this issue by tweaking this block of code from the [get_max_character_count](https://github.com/freelawproject/bigcases2/blob/1ad25e01276af345ab588a3d0312957e19ab27cb/bc/core/utils/images.py#L105-L143) method:

https://github.com/freelawproject/bigcases2/blob/1ad25e01276af345ab588a3d0312957e19ab27cb/bc/core/utils/images.py#L139-L140

We should also check the available space for the **title** using the if statement. like this:

```python
if self.get_available_space(wrapped_desc) <= 0 or self.get_available_space(wrapped_title) <= 0:
    break
```

The previous block of code will make sure the **description** and the **title** don't overflow the canvas before We update the `max_character` variable.

ERosendo:
This PR fixes #191 by tweaking the [get_max_character_count](https://github.com/freelawproject/bigcases2/blob/1ad25e01276af345ab588a3d0312957e19ab27cb/bc/core/utils/images.py#L105-L143) method and also adds tests to avoid getting this issue in the future.


bc/core/tests.py
```diff
@@ -94,7 +96,71 @@ def test_truncate_descriptions(self):
...
             self.assertTrue(
                 len(result) <= template.max_characters,
-                msg="Failed with dict: %s.\n"
-                "%s is longer than %s"
+                msg="Failed with dict: %s.\n%s is longer than %s"
                 % (test, result, template.max_characters),
             )
+
+
+class TextImageTest(TestCase):
```
mlissner, line 104: This should probably be a SimpleTestCase. They're faster if you don't need the DB.

bc/core/tests.py
```diff
@@ -94,7 +96,71 @@ def test_truncate_descriptions(self):
...
+                    "Transcript Order Form: re 115 Notice of Appeal,,,,"
+                    " transcript not requested Reminder: If the transcript is"
+                    " ordered for an appeal, Appellant must also file a copy"
+                    " of the order form with the appeals court. (Lynch,"
+                    " Christopher) (Entered: 04/10/2023)"
+                ),
+            },
+        )
+
+        for test in self.tests + EDGE_CASE:
```
mlissner, line 143: Do you need to run the `self.tests` here too? If so, it feels like you should just combine these into one test b/c they're so similar. If not, then two tests is fine, but `self.test` should go into the `test_creates_canvas_smaller_than_max_width`.

mlissner (changes requested):
A couple test tweaks, please.

bc/core/tests.py
ERosendo: You're right! I'm going to use SimpleTestCase instead of TestCase.

mlissner (approved):

Issue #152, opened 2023-03-24
Establish database table to track sponsorship values


mlissner:
In #16, we will begin purchasing documents automatically, but before we can do that, we need to establish a budget for doing so and we need to make a model where we track the purchases that we make. 

This issue is to establish the models that we'll need for this feature. The general idea works like this:

1. FLP finds a sponsor or decides to sponsor the bot itself.

2. After actual money has been given to FLP, somebody goes into bots.law and makes an entry into a table that says, "Org XYZ has paid $ABC dollars to sponsor the bot."

4. The next time the bot needs a document, it purchases it using that budget if there's still money there.
    - If the money is coming from FLP, it provides a thumbnail saying so and seeks further sponsorship (details in issue #151)
    - If the money is coming from another org, it watermarks the thumbnails.

5. After a document is purchased, a row is added to that table noting the purchased document and reducing the value in the account.

In the end, I think this just needs one table that can serve as the ledger. Money goes in via step 2 and is removed via step 5. The total for each user determines who sponsors and purchases a document.

If the ledger is down to zero across all accounts, the doc isn't sponsored.

Two more bits to think about:

1. Does every sponsor get an account in bots.law? Who makes it? Do we ask that sponsors make it? That'd be OK, but if so, that'd imply that they should be able to contribute funds (and check their funds) via the website (suddenly we need stripe, email, etc.). This is getting complicated.

2. If we have a bunch of people with money in their accounts, how do we determine which to spend? FLP's money should be spent as a last resort. We're the fallback. If we have two sponsors, we could:
    - round-robin their sponsorship (one, then the other, and back)
    - do weighted sponsorships (i.e., the one with 8× the account value sponsors 8/10 posts)
    - do first in first out (FIFO). 

    I think FIFO is best, because that way a big sponsor gets to have their name up non-stop for a week or month, instead of having it come and go for a while. I'm not sure how to easily implement this though using the table above. Hm.

mlissner:
Via: https://github.com/freelawproject/bigcases2/pull/165

Issue #24, opened 2023-01-26
Add system for trusted curators to add cases to channels


mlissner:
Curators should use the django permissions system. They should have the ability to add specific cases to specific channels (big cases, little cases, slack, discord, etc).

They could do this either via a dedicated webpage (and we create users for them via admin), or via the admin site, if we can control it enough. Eventually, we'll probably need a dedicated webpage.

mlissner:
Note that curators should automatically be listed on the website somewhere (see #22)

mlissner:
Feel free to list candidate curators here. I'll start with two:

 - Seamus Hughes: https://twitter.com/SeamusHughes
 - Chris Geidner: https://twitter.com/chrisgeidner

Obviously, the higher profile folks we get, the better.

mlissner:
Here are some others from another doc we've got:


[Dahlia Lithwick](https://slate.com/author/dahlia-lithwick), Slate & Amicus podcast
[Amy Howe](https://amylhowe.com/), SCOTUSBlog & Howe on the Court
[Avalon Zoppo](https://www.law.com/nationallawjournal/author/profile/avalon-zoppo/), National Law Journal
[Nina Totenberg](https://www.npr.org/people/2101289/nina-totenberg), NPR
[Linda Greenhouse](https://www.nytimes.com/column/linda-greenhouse), NY Times & Yale Law
[Katie Benner](https://www.nytimes.com/by/katie-benner), NY Times 
[Zoe Tillman](https://www.buzzfeed.com/zoetillman), Buzzfeed
[Ann E. Marimow](https://www.washingtonpost.com/people/ann-e-marimow/), Washington Post
[Maura Dolan](https://www.latimes.com/people/maura-dolan), LA Times
[Ginny Laroe](https://firstamendmentcoalition.org/2019/07/ginny-laroe-named-as-facs-programs-and-outreach-director/), First Amendment Coalition (former journalist)
[Leah Litman](https://michigan.law.umich.edu/faculty-and-scholarship/our-faculty/leah-litman), Univ of Michigan Law
[Jan Crawford Greenburg](https://en.wikipedia.org/wiki/Jan_Crawford), CBS News
[Joan Biskupic](https://en.wikipedia.org/wiki/Joan_Biskupic), CNN 
[Jessica Gresko](https://twitter.com/jessicagresko), Associated Press
[Marcy Wheeler](https://www.emptywheel.net/about/), @emptywheel
[...]

PR #205, opened 2023-04-19
feat(user): Enhance admin site for the User model

ERosendo:
This PR enhances the django admin interface for the `User` model by extending the `django.contrib.auth.admin.UserAdmin` class.

Django uses the `UserAdmin` class to render a nice-looking admin site for the `User` model. By just using this class in our `admin.py` file, we can get the same look for our model.  

The `UserAdmin` provides a custom **form** to create users from the admin panel(uses the [user/add_form.html ](https://github.com/django/django/blob/8b1ff0da4b162e87edebd94e61f2cd153e9e159d/django/contrib/admin/templates/admin/auth/user/add_form.html) file to render the form after clicking the `add user` button in the admin site) and We can customize the **fields** included in the form (like this PR does in line 12 of the admin.py file). This class will also take care of password validation and hashing. 

The form to add users from the admin panel looks like this:
 
image

However, this alone is not enough, since Django Admin will not display any of the customs fields from the User model when editing users, so We also need to tweak the `fieldsets` variable.

This enhancement will help us create users for curators(https://github.com/freelawproject/bigcases2/issues/24) and sponsors(https://github.com/freelawproject/bigcases2/issues/152) using a clean form in the admin site.

### DIFF
--- a/bc/users/admin.py
+++ b/bc/users/admin.py
@@ -1,5 +1,17 @@
 from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 
 from .models import User
 
-admin.site.register(User)
+
+class UserAdmin(BaseUserAdmin):
+    fieldsets = BaseUserAdmin.fieldsets + (  # type: ignore
+        # Custom fields added on to the bottom
+        ("Extra Fields", {"fields": ["affiliation"]}),
+    )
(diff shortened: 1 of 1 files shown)


mlissner:
Huh. I don't understand this code at all, but I'll take my chances and see how it appears. Thanks for the tweaks.

PR #210, opened 2023-04-24
feat(subscription): Add system for curators

ERosendo:
This PR adds a new page to add cases and fixes https://github.com/freelawproject/bigcases2/issues/24.

This new page has a search bar to lookup for cases and a web form allowing curators to tweak the case name and add a summary. 

Here's a gif showing this new page:

search-bar


bc/subscription/tests.py
```diff
@@ -0,0 +1,57 @@
...
+                "query": "https://www.courtlistener.com/docket/65680829/7/cornet-v-twitter-inc/",
+                "docket_id": 65680829,
+            },
+        )
+        for test in test_inputs:
+            result = get_docket_id_from_query(test["query"])
+            self.assertEqual(
+                result,
+                test["docket_id"],
+                msg=f"Failed with dict: {test}.",
```
mlissner, line 56: The assertEqual method does a really good diff, showing what you got vs what you were supposed to get. I don't usually override it unless I'm sure my message is better. In most cases, it's not, and I'd argue that this is a good example where your message isn't as good as the default. Maybe just remove this line?

bc/assets/static-global/js/htmx.min.js
mlissner, line 1: If there's a non-minified version of this, we should supply it to our developers, with something like:

```
{% if debug (or whatever %}
src=non-minified-version.js
{% else %}
src=this-version.min.js
{% endif %}
```

bc/assets/static-global/js/loading-states.js
mlissner, line 1: No minified version of this? Also the note above about the if/else for the minified version applies.

bc/assets/templates/base.html
```diff
@@ -19,6 +19,9 @@
     <link rel="icon" href="{% static "svg/icon.svg" %}" type="image/svg+xml">
     <link rel="apple-touch-icon" href="{% static "png/apple-touch-icon.png" %}">
 
+    {% block head_script %}
+    {% endblock %}
+
```
mlissner, line 24: I thought these days it was preferred to put JS scripts after the HTML so they load after the HTML has been downloaded and displayed? This could be obsolete info when using a CDN over HTTP/2. Do you know?

bc/subscription/templates/add-case.html
```diff
@@ -0,0 +1,23 @@
+{% extends "post.html" %}
+{% load static %}
+
+{% block head_script %}
+
+<script src="{% static "js/htmx.min.js" %}"></script>
+<script src="{% static "js/loading-states.js" %}"></script>
+
+{% endblock %}
```
mlissner, line 9: Nit:

```suggestion
{% block head_script %}
  <script src="{% static "js/htmx.min.js" %}"></script>
  <script src="{% static "js/loading-states.js" %}"></script>
{% endblock %}
```

bc/subscription/templates/includes/search_htmx/search-form.html
```diff
@@ -0,0 +1,29 @@
...
+  <div class="flex flex-col">
+    <div class="flex item-center">
+      <label for="simple-search" class="sr-only">Search</label>
+      <div class="relative w-full">
+        <div class="absolute inset-y-0 left-0 flex items-center pl-3 pointer-events-none text-gray-500">
+          {% include 'includes/inlines/magnifying-glass.svg'%}
+        </div>
+        <input type="text" id="simple-search" name="q"
+          class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-blue-500 focus:border-blue-500 block w-full pl-10 p-2.5 placeholder-gray-400"
+          placeholder="i.e. https://www.courtlistener.com/docket/65895581/carroll-v-trump/" required>
```
mlissner, line 11: Let's use this one, which is about ending PACER fees: 
```suggestion
          placeholder="i.e. https://www.courtlistener.com/docket/4214664/national-veterans-legal-services-program-v-united-states/" required>
```

mlissner (changes requested):
Looks good. I'm excited to use this. I think you've got the right direction, but I do have a handful of comments, as always. 

The one thing that this will finally force us to deal with is our CDN. I think I managed to reconfigure it properly, but we're in new territory, so we'll have to watch closely to make sure that the "Add Case"/"Log in" button is always correct once this is deployed.

Good stuff though, thank you.

mlissner:
By the way, the next thing we're going to want here is a search interface. I can see it coming and I'm sorry.

bc/assets/static-global/js/loading-states.js
ERosendo: There's still no minified version for this extension. Here's an issue you created about minified versions in the HTMX repo:

https://github.com/bigskysoftware/htmx/issues/1089

bc/assets/templates/base.html
ERosendo: You're right about the script tag, I think most of the JS we use these days can load after the web page is rendered but there are still scripts that advise placing the tag within the `<head> … </head>` because they're critical to the page contents and should be loaded first such as analytics or authentication-type of scripts, that is not our case so I'll move this block into the  `<body> … </body>`.

bc/subscription/utils/courtlistener.py
```diff
@@ -1,14 +1,21 @@
...
 from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.core.validators import URLValidator
 
 from .exceptions import MultiDefendantCaseError
 
 logger = logging.getLogger(__name__)
 
+DOCKET_URL_PATTERN = re.compile(
+    r"(?:www\.courtlistener\.com\/docket\/)(?P<docket_id>\d+)(?:\/.*)"
```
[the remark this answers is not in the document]
ERosendo: Sure!

bc/assets/static-global/js/htmx.min.js
ERosendo: You're right. We can use the [debug ](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#debug)template tag. I'm going to check the [INTERNAL_IPS ](https://docs.djangoproject.com/en/4.1/ref/settings/#internal-ips) list.

bc/subscription/views.py
```diff
@@ -0,0 +1,88 @@
...
+
+                queue.enqueue(
+                    api.add_status,
+                    message,
+                    None,
+                    retry=Retry(
+                        max=settings.RQ_MAX_NUMBER_OF_RETRIES,
+                        interval=settings.RQ_RETRY_INTERVAL,
+                    ),
+                )
```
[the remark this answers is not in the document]
ERosendo: You're right about the loop. We're using the same loop in a few functions. I'm adding a helper method to refactor this. 

The `message` should be generated inside the loop because It might depend on the channel. When this PR was created, the Bot was using the same template to create a status after following a new case in Mastodon and Twitter, We could've removed the message from the loop but it's not like that anymore because We created a custom template for Mastodon(with hashtags) in #213 so We need to keep the message inside the loop.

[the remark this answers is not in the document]
mlissner: Right, good point, thanks.

bc/assets/static-global/js/loading-states.js
mlissner: Ha, right, OK. My poor memory...

bc/assets/templates/includes/header.html
```diff
@@ -127,6 +127,11 @@
                         <a href="{% url 'collaboration' %}" aria-label="Chat Apps" class="text-sm lg:text-base font-medium uppercase hover:text-gray-700">
                             Chat Apps
                         </a>
+                        {% if user.is_authenticated %}
```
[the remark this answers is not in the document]
ERosendo: Yes, It makes sense to show the login button in the header for those users without the `sessionid` cookie. 

The header would look like this if We add the button to login:

image

Let me know what you think about the `[Sign in]` button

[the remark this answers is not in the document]
mlissner: LGTM. :shipit:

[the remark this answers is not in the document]
ERosendo: We can use the include from #214 to add the `Sign In` button and avoid repeating HTML code. We just need to add another variable to tweak the color of the element. We could do something like:

```
If We want to render a yellow button:
{% include 'includes/action-button.html' with link="https://free.law/contact/" text='Contact Us' size="sm" color='saffron' %}

If We want to render a white button:
{% include 'includes/action-button.html' with link="https://free.law/contact/" text='Contact Us' size="sm" color='white' %}
```

bc/subscription/tests.py
ERosendo: Sure. I will remove the custom message.

bc/assets/templates/includes/header.html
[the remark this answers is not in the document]
mlissner: Sure.

bc/subscription/utils/courtlistener.py
```diff
@@ -37,6 +44,34 @@ def map_cl_to_pacer_id(cl_id):
     return cl_to_pacer_ids.get(cl_id, cl_id)
 
 
+def get_docket_id_from_query(query: str) -> int:
```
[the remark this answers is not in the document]
ERosendo: I've added support for links like this one. We're going to use the API redirection to get the docket url.

bc/subscription/utils/courtlistener.py
```diff
@@ -64,6 +74,13 @@ def get_docket_id_from_query(query: str) -> int:
     validator = URLValidator()
     validator(cleaned_str)
 
+    # check if the query string is a PDF link
+    is_pdf_link = re.search(PDF_URL_PATTERN, cleaned_str)
+    if is_pdf_link:
+        r = requests.get(is_pdf_link.group("url_for_redirect"))
```
mlissner, line 80: Don't forget a timeout on all requests. https://github.com/psf/requests/issues/3070.

```suggestion
        url = is_pdf_link.group("url_for_redirect")
        r = requests.get(url, timeout=10)
```

pyproject.toml
```diff
@@ -50,6 +50,7 @@ django-tailwind = "^3.5.0"
 django-ratelimit = "^4.0.0"
 twitterapi = "^2.8.2"
 django-htmx = "^1.14.0"
+django-ses = "^3.4.1"
```
mlissner, line 53: Hm, should this be here?

ARCHITECTURE.md
```diff
@@ -34,6 +34,12 @@ To set up a development machine, do the following:
 
     This is important so that each service in the compose file can have a hostname.
 
+1. Get the Gateway IP of the new bridge network and paste it in the settings file:
+
+    `docker network inspect bc2_net_overlay --format "{{range .IPAM.Config}}{{.Gateway}}{{end}}"`
+
+    This is important so We can use the `debug` tag in the templates.
```
mlissner, line 41: I'm not familiar with this. Is there a reference you could share or that we could link to here?

bc/settings/django.py
```diff
@@ -45,8 +45,6 @@
 
 TAILWIND_APP_NAME = "bc.web"
 
-INTERNAL_IPS = ("127.0.0.1",)
```
mlissner, line 47: Is this no longer set when `DEVELOPMENT is False`? It looks like it's still set when `DEVELOPMENT is True`, but not when it's `False`?

mlissner (changes requested):
Almost there. 

Wow there were a lot of tune ups when you added the enqueing for loops!

I need some help understanding the gateway stuff, and one small tweak, I think. 

Thank you!

bc/subscription/utils/courtlistener.py
ERosendo: You're right! I will add the timeout.

pyproject.toml
ERosendo: No. Sorry, I  messed up the order of the dependencies when I fixed the merge conflict with the `main` branch. `django-ses` should be in line 52 and `django-htmx` in line 53.

ARCHITECTURE.md
ERosendo: I'm going to try to summarize what I learn from a few articles about the `debug` tag, the django `INTERNAL_IPS` setting, and the docker documentation about bridge networks.

In this PR, We include `HTMX` to the HTML templates, and We'd like to use the **minified** version for production and also supply  the non-minified versions to our developers(if they're available), so you suggested We can do something like: 

``` python
{% if debug %}
    src=non-minified-version.js
{% else %}
     src=this-version.min.js
{% endif %}
```

The previous `if statement` seems pretty easy to implement since django has a tag called `debug`, but after checking the docs I found out We also need to set the `INTERNAL_IPS` to let Django know that its ok to disclose sensitive information( like the `debug` variable) to the template's context. 

I checked the `setting` module and realized the `INTERNAL_IPS` variable was defined in the security.py file. We added just `127.0.0.1` to the list, but the `if statement` was still not working. Something was wrong with the list. I noticed some of the projects that suggest values for `INTERNAL_IPS`(like [django-tailwind](https://github.com/timonweb/django-tailwind)) didn't use Docker to set up the development environment, so I started looking at the Docker docs about networking.

I found that: 

> By default, a container gets an IP address for every Docker network it attaches to. A container receives an IP address out of the IP pool of the network it attaches to and each network also has a default subnet mask and gateway.

so the request originating from the `localhost` is passed through the **Gateway** of the network, due to which the Django application resolves the originating IP address of the request as the **Gateway IP** of the Docker network, this means that even if we have the set ["127.0.0.1"] as `INTERNAL_IPS` during development in the local machine, the django app will not add the debug variable to the context because the **Gateway IP** is not included in the list.

ERosendo: While I was looking for the links I reviewed to come up with this, I found that [django-debug-toolbar](https://github.com/jazzband/django-debug-toolbar) and [cookiecutter-django](https://github.com/cookiecutter/cookiecutter-django) implements a simpler and easier way to get the **Gateway IP** using the socket module.

They have the following code in the setting file:

```
import socket

hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
INTERNAL_IPS += [".".join(ip.split(".")[:-1] + ["1"]) for ip in ips]
```

in the previous code, we have a few operations: 

1. `socket.gethostbyname_ex(socket.gethostname())` returns the IP address of the host machine in the ips variable.

2. `[ip[:-1] + '1' for ip in ips]` removes the last digit of the host IP address and replaces it with 1. This is done because the host container receives its IP address out of the IP pool of the network and the pool usually uses the X.X.X.1 address as a default gateway (We could change the default gateway but that's not our case, We don't use the advanced options in the `docker network create` command). let's say the IP address of the host is 172.17.0.3 by default, the gateway of the Docker network is set 172.17.0.1, hence the last value of the IP address can be replaced with 1 to get the default gateway.

We can use this code and remove the extra step for developers..

ERosendo: let me know what you think.

mlissner: Well, that explains why I haven't been able to get the debug toolbar to work for a while. Let's do their approach and shall we install the debug toolbar too? We should probably get it working in CL again, if you're down in this area of thinking and it's easy to do.

poetry.lock
mlissner, line 1: You're adding a bunch of upgrades here that I don't think you intend, sorry. :)

ERosendo:
@mlissner thanks for your comments. I fixed the poetry.lock file, tweak the logic to get the Gateway IP, and added the debug toolbar as a dev dependency. 

I checked the toolbar docs to update the `settings` module and the `URLs` file. I found out We needed it to add [this script](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#htmx) in pages that use HTMX.

mlissner (approved):
There it is. Merging!

mlissner:
Will you see if a quick PR is possible for CL to add the debug toolbar there too?

ERosendo:
Sure!

PR #212, opened 2023-04-25
feat(core): Handle mastodon exceptions

ERosendo:
This commit adds logic to handle the `MastodonServerError` exception and the exceptions related to timeouts. 

This PR fixes https://github.com/freelawproject/bigcases2/issues/167 and closes https://github.com/freelawproject/bigcases2/issues/134 by trying to create a text-only post when the endpoint to upload media is failing.

This PR also changes the [default timeout](https://mastodonpy.readthedocs.io/en/1.2.1/#mastodon.Mastodon.__init__) (300 seconds) [mastodon.py](https://github.com/halcy/Mastodon.py) is using for all the request because it's greater than the timeout used by RQ (180 seconds), so tasks exceed maximum timeout value before We can catch and handle exception from the mastodon.py like in https://github.com/freelawproject/bigcases2/issues/197.


mlissner:
An OK solution to a not-great problem. Cool. When we have `text_image`s, those wont' get posted, and we'll still have a `[More info below]` thing, but I guess that's OK.

PR #214, opened 2023-04-27
feat(templates): Make buttons into an include

### LINKED ISSUE #143: Make buttons into an include
The buttons on BCB are repeated constantly and it's really hard to make one without doing a copy/paste.

Let's pull them into their own include so this isn't the case anymore.

ERosendo:
This PR adds an include for yellow buttons and one for transparent buttons. This PR fixes #143


mlissner (commented):
This looks pretty good, but it suffers from tailwind's big problem: Reviewing classes is really hard. I'm a little concerned about inconsistency in the buttons. Could you help me read this by pulling together a little info about which extra classes are in the various `{{ class }}` fields? Maybe we can drop those or move them to the default class list?

ERosendo:
I added the `{{ class }}` variable in the include to tweak the **padding** and the **responsive behavior**(**visibility** and **size**) of the **button** because We use this include a few times on templates like the **homepage**(the `Donate now` and `Join the waitlist` buttons are examples of this), the **header** and the **sponsor's page**, but each of the buttons uses different padding classes and has different responsive behavior.

While I was extracting the extra classes to create a summary for you,  I realized We could move some of them(the ones related to **responsive behavior**) to the parent element, so I updated the templates and cleaned the `{{ class }}` variable in the `include` tags.

Here's the list of tailwind extra classes We're using now:

| Button  | Extra classes | Info |
| ------------- | ------------- | ------------- |
| Donate (header) | px-6 py-2  | <p> padding-left: 1.5rem; <p> padding-right: 1.5rem; <p> padding-top: 0.5rem; <p> padding-bottom: 0.5rem;|
| Join the Waitlist (homepage)  | p-3  | padding: 0.75rem |
| Donate Now (homepage)   | p-3  | padding: 0.75rem |
| Contact Us (sponsors page)  | p-2  | padding: 0.5rem |

mlissner:
Hm, "Join the Waitlist" has `px-12 py-3`, so we probably don't want to just switch to `p-3`, right, or am I missing something? 

It'd be nice if we could normalize on a few sizes, so instead of doing `class="py-3"` or whatever, we could do `size="lg"`. I'm still not sure how feasible that is. It seems like the donate header button could become a `p-2`, and then we'd have:

`lg` -> `px-12 py-3`
`md` -> `p-3`
`sm` -> `p-2` (or maybe `px-4 py-2`

That'd be better than trying to remember and keep everything consistent, no?

ERosendo:
The `Join the Waitlist` button needed the `px-12` class to reach the desired width because It shared the same parent element with the text above it(a `div `tag), so there was no way to set the width of the button(using tailwind classes) from the parent without affecting the text, that's why I used the **padding** and the **size** of the button to make it responsive and fit the design but doing that added a bunch of utility classes to the `<a>` tag. 

After cleaning the includes, that button is wrapped by a div tag, so We can just set the **width** of this new parent `div`, let the button inherit it, and remove the `px-12` class. We still need the vertical padding because the border of the default button is too close to the text.

We can use the `p-2` class in the donate button in the header, and I think using a `size` variable to keep everything consistent is a fantastic idea.

mlissner (approved):

PR #216, opened 2023-04-27
feat(settings): Setup django-ses

### LINKED ISSUE #199: Set up AWS SES to send user-related emails
This shouldn't be too hard, but we'll have to get all the security stuff right. There are two parts to this:

1. Django needs to be set up with django-ses. This can just be copied from CL.

3. AWS needs to be set up to send emails from bots.law. 

Neither of these is particularly hard. I'll wait until the django-ses part is set up first, then I can do the SES stuff.

ERosendo:
This PR adds [django-ses](https://github.com/django-ses/django-ses) as a dependency and updates the `settings` module. This PR fixes #199.


mlissner (changes requested):
This looks fine, but for least privilege, we create separate accounts in AWS with email sending permissions. Therefore, we need something like this in aws.py:

```
# SES
AWS_SES_ACCESS_KEY_ID = env("AWS_SES_ACCESS_KEY_ID", default="")
AWS_SES_SECRET_ACCESS_KEY = env("AWS_SES_SECRET_ACCESS_KEY", default="")
```

ERosendo:
Thanks for your comments. I added the separate access key in the `aws.py` file.

mlissner (approved):

mlissner:
I created a new IAM user with SES access and set up and deployed the credentials. I had to also set up a custom `MAIL FROM` domain too, which required more DNS entries. Once that was done, this worked:

```
send_mail(
    subject=f"Testing  email for this moment",
    message="This is a test",
    from_email="noreply@bots.law",
    recipient_list=["mike@free.law"],
)
```

image

PR #231, opened 2023-05-02
fix(template): Remove inline elements from the add case template

ERosendo:
This PR fixes the errors with the CSP policies of the website caused by the inline elements in the `add-case` template.

- Adds a meta tag to tweak the configuration options of HTMX and avoid injecting inline styles. I used the comments from [CSP issue: style-insertion]( https://github.com/bigskysoftware/htmx/issues/862) to find the option that We should change. 
- Move the js code suggested on the `debug toolbar` docs into an asset to avoid inline scripts in the `add-case.html` template.


mlissner:
Wonky, but OK.

PR #235, opened 2023-05-04
feat(db): link subscriptions sponsorships and users

ERosendo:
This PR is a small part of the changes We're adding to support little case bots. 

- Adds the following tables and relationships to the database: 
     - A new table called `Alias`. This new table allows us to link channels together.
     - A new m2m table between `sponsorships` and `alias`.
     - A new m2m table between `subscriptions` and `channels`.
    

- Enhance the admin site for `Channels` and register the new `Alias` model. 
     - Introduces an inline to list channel's curators in the admin interface for this model.
Channel admin

     - This PR adds an inline for the Alias interface too.
Aliases

- Adds a list of channels(using an inline) to the subscription interface:
image

- Adds an inline to link `Aliases` to `Sponsorships` records,
image


bc/channel/models.py
```diff
@@ -1,13 +1,36 @@
...
+        help_text="Name for a set of channels",
+        max_length=100,
+    )
+    is_big_cases = models.BooleanField(
+        help_text="Designates whether this alias should be treated as the group of big cases channels",
+        default=False,
+    )
+    sponsorships = models.ManyToManyField(
+        Sponsorship,
+        related_name="channel_groups",
```
mlissner, line 23: When a related name doesn't match the name of the class, that's usually a smell. It usually means that you didn't like the name of the model and are thinking of it a different way.

I think in this case, calling these objects aliases doesn't really make sense. An alias a different name for something. These really are more like channel groups, right?

bc/channel/models.py
```diff
@@ -44,6 +67,9 @@ class Channel(AbstractDateTimeModel):
     enabled = models.BooleanField(
         help_text="Disabled by default; must enable manually", default=False
     )
+    alias = models.ForeignKey(
```
mlissner, line 70: Per my last comment, maybe this should be:

```suggestion
    group = models.ForeignKey(
```

?

bc/subscription/models.py
```diff
@@ -61,6 +62,12 @@ class Subscription(AbstractDateTimeModel):
         max_length=100,
         blank=True,
     )
+    channel = models.ManyToManyField(
+        Channel,
+        help_text="List of channels where a post related to this docket should be created",
+        related_name="Subscription",
+        blank=True,
+    )
```
mlissner, line 70:
```suggestion
    channel = models.ManyToManyField(
        Channel,
        help_text="Channels where a post related to this docket should be created",
        related_name="subscriptions",
        blank=True,
    )
```

mlissner (changes requested):
A few suggestions. You didn't want to put the user-related migrations in here?

bc/channel/models.py
ERosendo: Yes. I will update the name of this field.

ERosendo:
Thanks for your comments. I've reviewed and applied all the requested changes. 

I realized We already added an m2m relationship between `channels` and `users`. We included the user field in the channel model in #21. Sorry about this; I should've included this comment in the main description

https://github.com/freelawproject/bigcases2/blob/d104c97beac35fdb89778ad6152755f9b72244f3/bc/channel/models.py#L38-L43

mlissner (approved):
Cool. I have a feeling we're going to wish they were called ChannelGroups, but we can see how it goes.

Merging.

PR #236, opened 2023-05-08
feat(web): Tweak logic related to subscriptions. 

ERosendo:
This PR tweaks queries related to subscription and channels to use the relationships added in #235 and also update the page about the Big Cases Bot and the form to follow new cases.

- Adds a selector called `get_subscriptions_for_big_casesto` to filter the list of subscriptions and show only those linked to channels related to the Big Cases bot and not for the other new bots.

- Add helpers to show the list of channels as a table in the `post` and the `lookup` command

- Refactor the `post` command.

- Tweaks the `lookup` command to show a list of channels and add them to the new subscriptions. 

- Add a new selector function to get channels linked to a user

- Updates the form to add channels from the website to show the list of channels using checkboxes

- Adds a new JS asset to add interaction to the checkboxes in the new form

- Updates the search view to send an HTMX trigger event

- Tweak views in the subscription module to handle the new form to add cases. 

- Add factories for the channel, groups, subscriptions, and user model.

- Add tests for selectors in the subscription and channel module

Here's a gif of the new form to add cases:

add-case-form-new

### DIFF
--- a/bc/channel/management/commands/post.py
+++ b/bc/channel/management/commands/post.py
@@ -15,6 +15,18 @@ queue = get_queue("default")
 def handle_post_command(
     ids: list[int], mapping: dict[int, Channel | Group], text: str
 ) -> None:
+    """
+    Enqueues a job to create a post for each channel in the list of ids provided
+    by the user.
+
+    Args:
+        ids (list[int]): list of the id from the user's input.
+        mapping (dict[int, Channel  |  Group]): Mapping of objects in the table.
+        text (str): Text to include in the posts.
+
+    Raises:
+        ValueError: if the provided input is not in the mapping variable.
+    """
     for record_id in ids:
         record = mapping.get(record_id)
 
--- a/bc/channel/management/commands/post.py
+++ b/bc/channel/management/commands/post.py
@@ -1,48 +1,82 @@
+from django.conf import settings
 from django.core.management.base import BaseCommand
-from prettytable import PrettyTable
+from django_rq.queues import get_queue
+from rq import Retry
 
-from bc.channel.models import Channel
-from bc.channel.utils.connectors.masto import MastodonConnector
-from bc.channel.utils.connectors.twitter import TwitterConnector
+from bc.channel.models import Channel, Group
+from bc.core.utils.commands import (
+    show_all_channels_table,
+    show_channel_groups_table,
+)
+
+queue = get_queue("default")
+
+
+def handle_post_command(
+    ids: list[int], mapping: dict[int, Channel | Group], text: str
+) -> None:
+    for record_id in ids:
+        record = mapping.get(record_id)
+
+        if record is None:
+            raise ValueError(f"No channel {record_id}")
+
+        if isinstance(record, Group):
+            for channel in record.channels.all():
+                api = channel.get_api_wrapper()
+                queue.enqueue(
+                    api.add_status,
+                    text,
+                    None,
+                    retry=Retry(
+                        max=settings.RQ_MAX_NUMBER_OF_RETRIES,
+                        interval=settings.RQ_RETRY_INTERVAL,
+                    ),
+                )
+        else:
+            api = record.get_api_wrapper()
+            queue.enqueue(
+                api.add_status,
+                text,
+                None,
+                retry=Retry(
+                    max=settings.RQ_MAX_NUMBER_OF_RETRIES,
+                    interval=settings.RQ_RETRY_INTERVAL,
+                ),
+            )
 
 
 class Command(BaseCommand):
     help = "Post something manually to one or more channels."
 
-    def handle(self, *args, **options):
-        tab = PrettyTable()
-        channel_mapping = {}
-        tab.field_names = ["ID", "Service", "Account", "Enabled", "URL"]
-        db_channels = Channel.objects.all()
-        for channel in db_channels:
-            tab.add_row(
-                [
-                    channel.id,
-                    channel.service,
-                    channel.account,
-                    channel.enabled,
-                    channel.self_url(),
-                ]
-            )
-            channel_mapping[channel.id] = channel
-        self.stdout.write(self.style.SUCCESS(tab))
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--show_groups",
+            action="store_true",
+            help="Shows the list of groups instead of individual channels",
+        )
 
-        channel_input = input(
-            "Which channels? Input ID, comma-separate for multiple, "
+    def handle(self, *args, **options):
+        if not options["show_groups"]:
+            table, mapping = show_all_channels_table()
+        else:
+            table, mapping = show_channel_groups_table()
+
+        self.stdout.write(self.style.SUCCESS(table))
+
+        instance = "group" if options["show_groups"] else "channel"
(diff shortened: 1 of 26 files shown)


ERosendo:
Here are gifs of the new lookup command:

This is without the --show_groups argument 
new-lookup-add


This is using the --show_groups argument 
new-lookup-add-show-groups

bc/channel/management/commands/post.py
```diff
@@ -1,49 +1,95 @@
...
-                    channel.self_url(),
-                ]
-            )
-            channel_mapping[channel.id] = channel
-        self.stdout.write(self.style.SUCCESS(tab))
+        if not options["show_groups"]:
+            table, mapping = show_all_channels_table()
+        else:
+            table, mapping = show_channel_groups_table()
+
```
mlissner, line 76: I think the logic here should be flipped so the default is a grouped table and you have to use a special arg to get the *un*grouped results.

bc/channel/management/commands/post.py
```diff
@@ -1,49 +1,95 @@
...
+                api = channel.get_api_wrapper()
+                queue.enqueue(
+                    api.add_status,
+                    text,
+                    None,
+                    retry=Retry(
+                        max=settings.RQ_MAX_NUMBER_OF_RETRIES,
+                        interval=settings.RQ_RETRY_INTERVAL,
+                    ),
+                )
```
mlissner, line 47: Didn't we create a helper for this loop?

bc/channel/selectors.py
```diff
@@ -17,5 +18,47 @@ def get_mastodon_channel() -> Channel:
     return obj
 
 
-def get_enabled_channels() -> Iterable[Channel]:
+def get_all_enabled_channels() -> Iterable[Channel]:
```
mlissner, line 21:
```suggestion
def get_all_enabled_channels() -> QuerySet[Channel]:
```

bc/channel/selectors.py
```diff
@@ -17,5 +18,47 @@ def get_mastodon_channel() -> Channel:
...
+    """
+    Returns the set of all enabled channels.
+
+    Returns:
+        Iterable[Channel]: Set of channels
+    """
     return Channel.objects.filter(enabled=True).all()
+
+
+def get_channels_per_subscription(subscription_pk: int) -> Iterable[Channel]:
```
mlissner, line 31:
```suggestion
def get_channels_per_subscription(subscription_pk: int) -> QuerySet[Channel]:
```

bc/core/utils/tests/base.py
```diff
@@ -0,0 +1,3 @@
+from faker import Faker
+
+faker = Faker()
```
mlissner, line 3: Why do this? Why not just do `faker = Faker()` in your tests?

bc/subscription/management/commands/lookup.py
```diff
@@ -23,6 +57,11 @@ def add_arguments(self, parser):
             action="store_true",
             help="Save and Subscribe to the case found using the CL API",
         )
+        parser.add_argument(
+            "--show_groups",
+            action="store_true",
+            help="Shows the list of groups instead of individual channels",
+        )
```
mlissner, line 64: This feels backwards too, like it should be the default, not opt-in.

bc/subscription/management/commands/lookup.py
mlissner, line 1: It sort of feels like looking up cases isn't very useful. I haven't done it once. Should we rename this as "subscribe.py" and remove the `--add` option?

mlissner (changes requested):
This looks good. A couple little things. 

On the interface, I think you need to take another look at how the form looks. The checkbox on the group is a little unintuitive. Maybe use intending of the channel rows, or perhaps make the checkbox into something else on the group rows (like a button that says "all" or maybe right align it?) I'm not sure, but it needs some kind of fix to distinguish between the header rows (with groups), and the content rows (with channels).

bc/channel/selectors.py
ERosendo: Thanks for the suggestion. I'll change the type hint.

bc/subscription/management/commands/lookup.py
ERosendo: You're right about the command. I've always used it with the `--add` option.  I will rename it and tweak the logic to show the grouped table as default.

bc/channel/management/commands/post.py
ERosendo: thanks for your comment. I'll change the logic here.

bc/core/utils/tests/base.py
ERosendo: I found the recommendation in this [article](https://www.hacksoft.io/blog/improve-your-tests-django-fakes-and-factories) from the [django-guidestyle](https://github.com/HackSoftware/Django-Styleguide#how-to-ask-a-question-or-propose-something). it says: 

> We highly recommend "proxying" the Faker instance and using it in your app that way.
> 
> You'd most likely want to have the same configuration when you use fakes around your app. Same goes if you want to customize the providers and use them in different places.

I also checked the [Django-Styleguide-Example](https://github.com/HackSoftware/Django-Styleguide-Example) repo and They're still using this approach to create a [Faker](https://faker.readthedocs.io/en/master/index.html) instance.

bc/channel/management/commands/post.py
ERosendo: We included a helper to refactor a loop similar to this one but it's not the same (even though it iterates over the channel list). 

In the `enqueue_posts_for_new_case` helper, We use one of the templates to compute the content of the post(We don't need to provide a message), but the loop in the `post` command is different because We use the plain text provided by the user through the command and create a new status in each channel (We don't need to use templates or create thumbnails)

I'm going to add a helper to refactor this loop.

ERosendo:
Here are a few mocks for the interface:

- Using a [Select all / none] button
new-select-all-form

- Using a [check all] button 
new-form-only-check

mlissner:
The `All/None` seems best to me. Maybe take off the word "Select" though? I think that's implied and people will get it. 

Can we also add `autocomplete=off` to the search field?

ERosendo:
Sure! I will add the `autocomplete` attribute.

ERosendo:
@mlissner I have applied your suggestions. Thanks for your comments

bc/subscription/templates/includes/search_htmx/case-form.html
```diff
@@ -35,20 +35,21 @@
...
-                {{ group.name }}
-              </div>
-            </label>
+          <div class="p-2 bg-gray-50 flex justify-between font-medium">
+            <div>
+              {{ group.name }}
+            </div>
+            <div class="text-sm self-end">
+              <a href="javascript:void(0)" class="underline text-blue-600 hover:text-blue-800" data-cg-target="group-{{group.pk}}" data-cg-uncheck="check-none-group-{{group.pk}}">All</a> /
+              <a href="javascript:void(0)" class="underline text-blue-600 hover:text-blue-800" id="check-none-group-{{group.pk}}">None</a>
```
mlissner, line 44: Is there a reason to do these as links instead of as spans that look like links?

mlissner (approved):

PR #243, opened 2023-05-17
fix(subscription): Use the pk property of webhook_event argument

ERosendo:
This PR fixes the `enqueue_posts_for_docket_alert` helper after the updates We introduced in #236 

The current implementation of this helper is passing `pk` getter property of the `FilingWebhookEvent` class (in line 81) to the `enqueue` method, but We should've used the `webhook_event` argument instead (like in line 76). The Bot is not creating new statuses because the enqueue method is raising an exception when it tries to de-serialize the Python object we're passing as the pk. 

This PR updates the arguments used in the `enqueue` method to remove the class and use the `webhook_event` argument instead.

### DIFF
--- a/bc/subscription/tasks.py
+++ b/bc/subscription/tasks.py
@@ -65,7 +65,7 @@ def enqueue_posts_for_docket_alert(
     handling a docket alert webhook.
 
     Args:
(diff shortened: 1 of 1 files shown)


mlissner:
Thanks.

PR #247, opened 2023-05-23
feat(channel): Save channel's credentials in the database

ERosendo:


mlissner (approved):
This is an easy one. We'll merge when it's ready. 

For deployment, I think we should remember to delete things from our `.env` file.

bc/channel/models.py
```diff
@@ -67,13 +71,26 @@ class Channel(AbstractDateTimeModel):
     group = models.ForeignKey(
         "Group", related_name="channels", null=True, on_delete=models.SET_NULL
     )
+    access_token = models.TextField(
+        help_text="Access Tokens of the account that the bot is making the request on behalf of",
+        null=True,
+    )
+    access_token_secret = models.TextField(
+        help_text="Access Tokens Secret of the account that the bot is making the request on behalf of",
+        null=True,
```
mlissner, line 80: Should these be nullable instead of blankable?

ERosendo: Thanks for pointing out this. It should be blankable. The Django docs advise not to use `null` on string-based fields.

PR #257, opened 2023-06-01
feat(user): Adds fields to handle authentication.

ERosendo:
This PR adds three new fields to the User model. These fields are:

- Activation_key
- key_expires
- Email_confirmed

These fields allow us to improve the user authentication process.

This PR also tweaks the admin interface for the user model to show the `Email_confirmed` field. We need to set this field to `True` for existing users before We merge #258.


mlissner (approved):
LGTM, thank you.

PR #259, opened 2023-06-01
feat(subscription): Tweak the links to the CL dockets

### LINKED ISSUE #252: Docket links should be reverse chron
When we link to dockets, they're ordered chronologically. It'd be better if they were reverse cron so the user saw the most recent thing at the top.

This is easy to do by adding `order_by=desc` to the URL.

ERosendo:
This PR adds the `order_by` query parameter to the docket links and fixes #252.

This PR updates the property called `cl_url` in the `Subscription` model. The current implementation of this property computes links that redirect the users to the **docket page** in CL but We can't add query parameters to this URL because **Django** removes it while redirecting the request, so We're tweaking the property to return the direct link to the docket page and also adding the order_by parameter.

### DIFF
--- a/bc/subscription/models.py
+++ b/bc/subscription/models.py
@@ -81,7 +81,7 @@ class Subscription(AbstractDateTimeModel):
 
     @property
     def cl_url(self) -> str:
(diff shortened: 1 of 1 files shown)


mlissner:
Simple enough, thanks.

PR #274, opened 2023-06-19
feat(channels): better alt-text for thumbnails and images

### LINKED ISSUE #230: Improve alt text by providing document text
It'd be nice if our alt text were better. Currently it just says, "Thumbnail of page X of the PDF".

**First,** we could make that better just by saying, "Thumbnail of page X of the PDF linked above."

That makes it more clear that clicking the link will get you the text.

But **second**, we *have* the text in our database. Could we do better by including it in the post? I think the answer is yes, and I think the way to do it is to ask the recap-document API for the text and use it if it's available, skipping it if not (implying that OCR is still going, which we shouldn't wait for). 

When we get the text from the API, we won't know what text came from which page, so we'll just have to dump as much of it as we can along with each thumbnail, possibly with some explanatory text:

> The first XXX characters of the PDF: "xxxx…"

Then on the second thumbnail: 

> The second XXX characters of the PDF: "…yyy…"

Then on the third thumbnail: 

> The third XXX characters of the PDF: "…zzz…"

(Of course, this should break at word boundaries, not in the middle of words.)

The alt-text bot goes further and even will do an image that just says "ALT TEXT" with additional text on it, but I think we can stop short of that (we're not providing all the pages, after all): 

https://twitter.com/AltTextUtil/status/1653058214362238976

One open question is whether we'll want to pre-process the text to remove whitespace. I think we probably will, and I wonder if we'll want to go even further to remove dumb punctuation at the beginning too. Like, maybe all we want are the words? This will take some experimentation, I think.

weedySeaDragon:
This is the first step toward improving the alt-text for thumbnails and images.

- For thumbnails, the alt-text is: **"Thumbnail of page X of the PDF linked above."**
- For images, the alt-text is: **"An image of the entry's full text: {text_image_description}"**

   where the `text_image_description` is the image description, e.g. text_image.description

The alt-text is quite simple; it can be improved and expanded in later PRs.

### Implementation Notes:
- Added 2 functions for creating the alt-text that can be called by anything.
- Changed `MastodonConnector` and `TwitterConnector` to use the new alt-text functions
- Added tests to ensure that the correct alt-text was being created and used
   - since there were no test for TwitterConnector, I created a new test file and tests for `add_status()` and `upload_media()`
   - renamed `test_utils.py` to `test_mastodon_conn.py` since it was really just testing MastodonConnector. (I'm sure that it was written at the start of developing this, when the MastodonConnector was the only connector or when it was just a collection of utility functions.) 
   - MastodonConnector: added tests for `add_status()` and `upload_media()` that parallel the tests for TwitterConnector

I may have overdone it with the tests.  I wrote them in part so I could understand the code and to get up to speed on the python mocks. 

The tests might be too tightly coupled to their targets (MastodonConnector and TwitterConnector). This is in part a function of how the classes and methods are written (e.g. can't really test that the correct alt-text was generated because it's never exposed by anything any methods return).  It might also be that the tests are too detailed (perhaps I'm testing too much; checking the internal states a bit too much, etc.).

You can let me know what the right approach is for this project.

Since this is my first PR for the project, please let me know if anything needs to be changed: the commit messages, any comments I made, formatting, naming, I did too much, I did too little.... whatever.  

Fixes #230


mlissner:
This all sounds really great. @ERosendo can you do a review, please and see if there are any changes we'd normally ask for?

Thank you!

ERosendo:
This looks good to me! Thank you for writing all the tests and updating the channel factory.

weedySeaDragon:
Is there something else I need to do for this?
I _assume_ that one of the maintainers will merge it. (Not intending to be impatient. I know it'll be merged in due time. Just checking in to make sure I've done what's needed.)

mlissner (approved):

mlissner:
The delay was just me being on vacation the last two weeks. I'm back though! I just skimmed the code, since @ERosendo gave it a more careful look. Three little thoughts before merging:

 - You say this PR fixes #230 in your comment above. I wouldn't do that, since that'll auto-close it.
 - You'll also want to get pre-commit working locally, so that the pre-commit bot doesn't pollute your commit log.
 - I appreciate your detailed commits! I always think of them as my blog and encourage lots of detail, even though it'll only ever be read if there's a problem.

Issue #273, opened 2023-06-19
chore: add dummy data for big cases


weedySeaDragon:
When I first got the system up and running and went to http://localhost:8888/big-cases/about/, there were no big cases being followed. I assumed that the initial dummy data would at least have one big case being followed so that something would show up on this page. Having some cases on the big cases page would let people see how cases show up on the page and that they are connected to Court Listener.

Having richer data will make it easier for developers to see how the models and information is related.  (Will especially help developers new to the project.)

At a minimum, I think `bootstrap-dev` should also:
1. create a channel Group for big cases: `name = 'big cases', is_big_cases=True`
2. set the Mastodon channel to be in the 'big cases' Group
3. add 2 of the existing Subscriptions to the Mastodon channel. Having more than 1 case lets a developer play with more variations.

I'm happy to do this. (I've already done the above just so I could see it on the big-cases page.)

Is there other minimal data that should be added?
Perhaps a Twitter channel should also be added? (And it should also be set to the 'big cases' Group)

mlissner:
The broad strokes all sound great to me. Any thoughts about the details, @ERosendo?  You're the second developer on this, so that's why none of this stuff exists yet! 

Want to take a look at how we build dev data on CourtListener using Faker? I think something somewhat similar might make sense here? I think the command is make_dev_data?

weedySeaDragon:
>  Want to take a look at how we build dev data on CourtListener using Faker? I think something somewhat similar might make sense here? I think the command is make_dev_data?

I've started looking at that. :-)  (It is indeed `make_dev_data`)
CL uses Factories to generate data.  I like that -- it keeps data generation to just one source (instead of a source for generating "dummy" (seed) data, and a separate second source for generating testing data. oy.)

I'll explore this but will wait for @ERosendo to chime in with anything that I should know.

ERosendo:
These steps sound great!

weedySeaDragon:
### Warning: OT No reply is expected

Funnily enough (to me), the very day that I'm working on the simplified fake citation stuff, the judge in the 'lawyer that used ChatGPT' case (Mata v. Avianca) just issued their opinion. 

I saw it on Twitter from someone I follow and yes, she *did* post the opinion from Court Listener.

[Mata v. Avianca, Inc., 1:22-cv-01461, (S.D.N.Y. Jun 22, 2023) ECF No. 55](https://www.courtlistener.com/docket/63107798/55/mata-v-avianca-inc/)


[...]

PR #277, opened 2023-06-23
chore(fake data): Faker provider LegalCitationsProvider

weedySeaDragon:
This is a Faker provider to generate fake legal citation strings and the sub-strings that go into the citations.

This is a simplified version of [LegalProvider in CourtListener](https://github.com/freelawproject/courtlistener/blob/main/cl/tests/providers.py).  
- So that this is self-contained, this does not use any external sources such as `reporters_db` or `juriscraper`.  Future versions could add these as optional data sources
- instead of using the external method `oxford_join`, this has and uses `humanized_join`. See question below.
   - This is not meant to also work with django querysets, just strings.

In the future, this could be extracted so that it is a stand-alone module that multiple projects could use it. At that point, it'd be good to allow optional data sources (e.g. reporters_db, etc.)

for #273
### Questions/Issues:

1. Naming:
   1.  Is the class name ok? 
   1.  I named the method `case_party_name` because naming it `party_name` seemed ambiguous to me.  But in this context (legal citations), perhaps `party_name` is best (and not ambiguous at all).
1. `humanized_join`:
   1. Have I left out important functionality from `humanized_join()` ?  I wanted to keep that simple, but am not sure if I went too far.
1. `case_name` does not use `titlecase`. Should it?
1. Tests - I did not write tests for this, but can.
1. I had put in some hard-coded citations and set them to be used about 5% of the time. The hard coded citations I usedwere the bogus cases generated by ChatGPT in the Mata v. Avancia case.
I can put these back in or any other hard-coded citations that could be helpful.


mlissner:
>     * So that this is self-contained, this does not use any external sources such as `reporters_db` or `juriscraper`.  Future versions could add these as optional data sources

Great. Good call. 


>     * instead of using the external method `oxford_join`, this has and uses `humanized_join`. See question below.

I, um, don't actually think it uses `humanized_join`? If that's the case, should we pull it out of here? 

> In the future, this could be extracted so that it is a stand-alone module that multiple projects could use it. At that point, it'd be good to allow optional data sources (e.g. reporters_db, etc.)

Cool. I'm open to that, but I admit that maintaining all these repos has gotten pretty difficult.

>        1. Is the class name ok?

Well, it does more than just create citations, so I think the original name was better.

>        2. I named the method `case_party_name` because naming it `party_name` seemed ambiguous to me.  But in this context (legal citations), perhaps `party_name` is best (and not ambiguous at all).

I don't find `party_name` ambiguous. What else did it remind you of? I prefer `party_name`, but I'm open to being convinced non-legal folks would find it confusing.

 >     2. `humanized_join`:
>        
>        1. Have I left out important functionality from `humanized_join()` ?  I wanted to keep that simple, but am not sure if I went too far.

Seems fine, but buggy (comment forthcoming).

>     3. `case_name` does not use `titlecase`. Should it?

Nah, I think ti's fine. You could use Python's titlecase if you wanted, instead of using the fancy one in Juriscraper (or is it in CourtListener, I forget).

>     4. Tests - I did not write tests for this, but can.

Tests for fakers seems gratuitous to me.

>     5. I had put in some hard-coded citations and set them to be used about 5% of the time. The hard coded citations I usedwere the bogus cases generated by ChatGPT in the Mata v. Avancia case.
>        I can put these back in or any other hard-coded citations that could be helpful.

Ha. Put 'em if you want, but it's probably not worth the effort at this point. My last Easter Egg caused a bug...

bc/core/utils/faker_legal_citations.py
```diff
@@ -0,0 +1,264 @@
...
+        :rtype: str
+        """
+        office = random.randint(1, 7)
+        year = random.randint(0, 99)
+        letters = random.choice(["cv", "bk", "cr", "ms"])
+        number = random.randint(1, 200_000)
+        return f"{office}:{year:02}-{letters}-{number:05}"
+
+    @staticmethod
+    def humanized_join(
```
mlissner, line 217: I don't see where this is being used, but I think it's buggy wrt to how it handles spaces around conjunctions. I think as it is presently, it'd create double spaces:

> `1, 2,  and  3`

I'd suggest having the code handle adding the spaces properly, and not to ask the user to remember to do ` and `. That feels pretty wonky to me.

mlissner:
I added my notes. @ERosendo if you can shepherd this one through any changes that wind up being needed and get it merged once you guys are happy, that'd be wonderful. I don't think it needs more code review though (but maybe I'm wrong!).

weedySeaDragon:
Here's the summary of the changes I just pushed (commits obv. show more detail):
1. Fixed `humanized_join()`
   - should this just be renamed to `oxford_join`?  I originally thought this would be confusing if this is ever used in CourtListener, but that may be overthinking.
2. Added the `full` parameter to party_name() and case_name() -- the original version had this.  Now these methods call `humanized_join()`
1. Changed the class name to `LegalProvider` per the original.  I personally find this name ambiguous since it assumes that "Provider" is always associated with "Faker".  My preference would be `LegalFakerProvider`. 
1. Made all methods static.  This is definitely a style/culture choice. This allows the methods to be called directly from the class; Faker is not required. (I added an example in the docstring for the class.)

bc/core/utils/legal_provider.py
```diff
@@ -0,0 +1,331 @@
...
+        Ex: `LegalProvider.citation_with_case()`
+            returns "Hooper, Gardner and Perry v. Bond Group, 579 L. Ed. 102"
+
+        :returns: the fake citation
+        :rtype: str
+        """
+        return f"{LegalProvider.case_name()}, {LegalProvider.citation()}"
+
+    @staticmethod
+    def citation() -> str:
```
ERosendo, line 68: We don’t use citations in the bot. Is there a reason you’ve added it? If not, we should delete this method and the one that makes citations with case names.

ERosendo (commented):
This looks pretty good. Just one comment.

Thanks!

bc/core/utils/legal_provider.py
weedySeaDragon: Definitely can do that.

I included them because they were in the CL version, and with an eye towards having this available as something that can be used for both Big Cases and CL.
But they can obv be added later if/when that happens.

I'll delete them. (They can always be re-added by looking at a previous commit.)

mlissner: Thanks. I think if we/you want to do a little faker package for legal data, they'd make a ton of sense there. We're sort of philosophically (?) opposed to adding unused code though, under the YAGNI banner. I'm not sure it's the right thing, but it's *our* think, anyway...

ERosendo:
@mlissner I think we can merge this PR :shipit:

mlissner:
With pleasure! Thank you everybody.

PR #279, opened 2023-06-23
chore(factories): expand Group- and Channel- Factories

### LINKED ISSUE #273: chore: add dummy data for big cases

weedySeaDragon:
for #273

Add traits, documentation to GroupFactory and ChannelFactory. 
Also adds  `fake_token()` (used by ChannelFactory and potentially in tests)

Per #273, Expanded the factories so that they can be used to create some initial data for development:
- Created traits for commonly used factory configurations:
   - GroupFactory:  `big_cases` and `little_cases`
   - ChannelFactory: `mastodon` and `twitter`

- Added some sensible attribute defaults
- Added documentation

(Accidentally based this branch on the faker-citation-data-273 branch. Rebased onto main)


mlissner:
I've added @ERosendo as a reviewer, thanks for this helpful work.

ERosendo:
Looks good to me, but you've got a conflict and it needs a rebase, please (there are changes here from earlier PRs).

ERosendo (approved):
@weedySeaDragon thank you! 

@mlissner I think we can merge this PR :shipit:

mlissner:
Thanks all!

PR #280, opened 2023-06-24
feat(web): Re-do the Suggest-a-Bot form

### LINKED ISSUE #261: Re-do the Suggest-a-Bot
It currently has:

 - Name: Of the bot or the submitter? It's unclear.
 - Platform: This is OK.
 - Description: Not very clear what details we need.

The name field should be for the name of the bot, so we should just change to "Bot Name:"

We need more from the description field, so one thing we can do is just templatize it with some pre-filled questions (like Github Issue Templates, say):

 - Purpose of the bot:
 - Your full name
 - Your email address
 - Your expertise in this field
 - Would you want to curate the bot? If not, who would you recommend?

ERosendo:
This PR adds new fields to the form to suggest a bot and fixes #261.

- Adds the following fields to the BotSuggestion model:
    - `user_full_name`
    - `user_email`
    - `user_expertise`
    - `is_curator`
    - `suggested_curators`

- Tweaks the BotSuggestionForm class to show all the new fields

- Updates the suggestion-form.html file and adds a new .js file to introduce interactivity to this form

Here are some screenshots of the new form:

- **When the user wants to curate the bot**
image

- **It will show a new field at the bottom when the user doesn't want to curate the bot**
image


mlissner:
Looks good. Set for auto-merge.

PR #281, opened 2023-06-24
chore(factories) SubscriptionFactory - use LegalCitationProvider data

### LINKED ISSUE #273: chore: add dummy data for big cases

weedySeaDragon:
for #273

Expand the `SubscriptionFactory` to use fake data generated by the Faker  `LegalCitationProvider` class. Also adds a docString for the post_generation method `channels()`

Depends on the chore/faker-citation-data-273 branch. This is the branch that adds `LegalCitationProvider`


mlissner:
I've put this on @ERosendo's backlog for review in the next couple days. Thanks @weedySeaDragon.

bc/subscription/tests/factories.py
```diff
@@ -13,13 +18,25 @@ class Meta:
...
+        @see https://factoryboy.readthedocs.io/en/stable/reference.html?highlight=post_generation#post-generation-hooks
+        @example:
+            some_channel = ChannelFactory.create()
+            SubscriptionFactory(channels=[channel_1, channel_2]) #
+            subscribes this to channel_1 and channel_2
+        """
         if not create or not extracted:
             return
-
+        print(f"extracted: {extracted}")
```
ERosendo, line 37: I think we should remove this print statement.

weedySeaDragon: Oops! Yes, I definitely should.

ERosendo: Thanks!

PR #285, opened 2023-06-28
feat(alt text) shorten text for images

weedySeaDragon:
for #230 

When describing an image, screen readers will already announce that it's an image because of the  tag, so there's no need to also say that something is an image in the alt text.
The alt text for an image is changed
from An image of the entry's full text: {text_image_description}
to The entry's text: {text_image_description}

Also removes "full" so that the alt text is a wee bit shorter.


mlissner:
Nice one, thanks!

PR #286, opened 2023-06-29
feat(subscription): Add thumbnail of initial complaint 

### LINKED ISSUE #217: Add thumbnail of initial complaint to "I am now following" message
When we subscribe to cases, there's often not a lot of context. It'd be nice if we could include thumbnails from the initial complaint when we do so, creating something a lot more like our regular posts. 

We could even grab the description from the first entry, like we do the description of updated cases.

ERosendo:
This PR fixes #217.

- Adds a new field to the subscription model to store article links.

- Adds type hints and a docstring to the `get_template_for_channel` method,

- Adds a new form to create a subscription. This new form takes care of the URL validation. 

- Tweaks the `case-form` template to use the new Django form. 

- Add a new message template for new cases with links to an article and updates the `get_new_case_template` method.

- Adds the article URL to the method that `creates/updates` subscription records.

- Updates the method called `enqueue_posts_for_new_case` to use the new message template and the URL in the subscription record.

- Adds a helper function that uses CL API to look up initial complaints.

- Adds logic to purchase the initial complaint.

- Adds the `docket_id` to the fetch request. 
        
    - We need to add this `id` to the request so We can link the webhook to a subscription when the event is triggered by an initial complaint. 

- Adds new dataclass to represent documents

- Tweaks the helper method that logs purchases.
         
    - This PR removes the `filing_webhook` parameter from the `log_purchase` method and adds a new parameter called document. This change reduces the coupling of the code because the new approach allows us to use this helper to log purchases of initial complaints and documents from filing webhooks.

- Adds a new helper method called `check_initial_complaint_before_posting`. This new method uses the lookup helper to check if the initial complaint is in the RECAP archive and tries to download the file. This helper can also purchase files if there are sponsorships available.

- Adds test for the new methods and also tweaks the `Subscription` factory to create objects with URLs.

- Update tests for the log_purchase method

Here are some screenshots of the new form and the new message template:

- **New form with the URL input** 
Screenshot 2023-07-20 at 11 09 39 AM

- **Post with link**
Screenshot 2023-07-20 at 11 27 48 AM

- **Post with a link and thumbnails**
image

- **Post with a link and sponsored thumbnails**
image


bc/core/utils/status/templates.py
```diff
@@ -81,3 +93,14 @@
...
 #CL{docket_id}""",
 )
+
+TWITTER_FOLLOW_A_NEW_CASE_W_ARTICLE = TwitterTemplate(
+    link_placeholders=["article_url"],
+    str_template="""I'm now following {docket}:
+
+{docket_link}
+
+[read more: {article_url}]
```
mlissner, line 103:
```suggestion
Docket: {docket_link}

Context: {article_url}
```

mlissner (approved):
Wow, lots of work here. I think this should have been two PRs, one for adding thumbnails and the other for doing the article link, but that's OK. I really don't have any worries on this one. Everything looked good to me with the exception of a couple of tweaks I already shared. 

If you want to make those tweaks and then merge, I'd say go for it! 

Thanks for all the work on this one.

ERosendo:
Here are some tests with the new template:

**- Post with link**

Screenshot 2023-07-25 at 9 26 34 AM

**- Post with a link and thumbnails**

Screenshot 2023-07-25 at 9 23 59 AM

PR #296, opened 2023-07-03
chore(deps): configure Dependabot to create grouped updates

ERosendo:
This PR tweaks the `dependabot.yml` file to group version updates by dependency name.


mlissner:
Looks like a nice start, thanks!

Issue #287, opened 2023-06-30
Add Sentry CSP Report URI


mlissner:
Sentry has a neat tool that can receive CSP reports when they fail. We're turning it on in CL and we should do it here too.

@cweider is working on this. The report URI for bots.law is: https://o399720.ingest.sentry.io/api/4504573893738496/security/?sentry_key=55e2a29e6b964e768eeebf0a368af7cd

Thank you!

PR #302, opened 2023-07-04
feat(security): add CSP violation reporting

cweider:
Sentry.io provides a CSP violation reporting endpoint; we should use it. This change adds the environment variable, `SENTRY_REPORT_URI`. If present, it is handed to `django-csp` and used for the `report-uri` directive.

Fixes: #287


mlissner:
Looks great, thank you. Let's see how this goes!

Issue #297, opened 2023-07-03
Tailwind checkbox image violates CSP


cweider:
<img width="1735" alt="Screenshot 2023-07-03 at 1 01 10 PM" src="https://github.com/freelawproject/bigcases2/assets/45040/32075e09-58a4-4b36-b7d4-fc1ad5c62d8c">

mlissner:
Perfect

cweider:
This is caused by `@tailwindcss/forms` plugin, which [uses data URIs for form elements in CSS](https://github.com/tailwindlabs/tailwindcss-forms/blob/master/src/index.js#L228). Fix is to except `data:` for `image-src`. This is not great, because `data:` can, literally, be anything, but images have limited scope for bad behavior and so, it’s not a pressing concern.

mlissner:
Hm I just tried to trigger this so we could test the new CSP reporting, but couldn't do so. At least just loading the page wasn't enough in Chrome...

ERosendo:
@mlissner You have to click the checkbox to trigger the error. :)

Here's the issue from Sentry:

[Blocked 'image' from 'data:'](https://freelawproject.sentry.io/issues/4296050084/?project=4504573893738496&query=is%3Aunresolved&referrer=issue-stream&stream_index=0)

mlissner:
Got it. Cool!

PR #315, opened 2023-07-05
fix(security): adjust CSP policy to accomodate Tailwindcss

cweider:
The `@tailwindcss/forms` plugin, uses data URIs for form elements in CSS. Allow these by adding `data:` to the allowed image sources of the content security policy.

Fixes: #297


mlissner:
Thank you!

Issue #314, opened 2023-07-05
Z-Index isues in drop down menu


mlissner:
STR:
 - Make your screen narrow to view mobile menu
 - Load this page: https://bots.law/reset-password/
 - Note how the email box is somehow on top of the menu:

![image](https://github.com/freelawproject/bigcases2/assets/236970/1c6a1184-a115-4c28-a6d8-f41ce8b1bc3d)

PR #317, opened 2023-07-06
fix(web): z-ordering of dropdown

cweider:
`z-index` determines ordering relative to an element's closest `relative` ancestor. The z-order of the dropdown is relative to z-order of its `absolute`-ly positioned parent.

This parent and the container of the input share a `relative` container (`<body>`) and, having no explicit index, are ordered by lexically. The input container comes later, and so it is ordered above the dropdown container. A child element cannot transcend its parent's z-order, and so the input rendered above the container. Fix by placing the `z-index` on the `absolute` container.

Fixes: #314

### DIFF
--- a/bc/assets/templates/includes/header.html
+++ b/bc/assets/templates/includes/header.html
@@ -15,9 +15,9 @@
                       {% include './inlines/bars.svg' %}
                     </div>
                 </button>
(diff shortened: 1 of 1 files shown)


ERosendo:
This looks good, Thanks @cweider!

@mlissner we can merge this PR :shipit:

PR #325, opened 2023-07-07
feat(security): add `report-sample` to script-src of CSP

cweider:
This keyword leads the browser to add a `sample` entry to violation reports (for `style` and `script`) and is useful for distinguishing otherwise identical reports. It is also helpful for diagnosis.

The sample is 40 characters of context for the source code that triggered the violation. It does not contain content (read: PII) (https://www.w3.org/TR/CSP3/#violation-sample).


mlissner:
What an amazing specification this is. I'm kind of impressed.

PR #329, opened 2023-07-20
chore(security): tighten-up CSP configuration

cweider:
Courtlistener recieved more recent attention to its CSP directives. Now that the policy has been dialed- in there, bring these changes to Big Cases.

- Consistently format `*_SRC` lists
- Consistently specify protocol for sources (see `AWS_S3`)
- Specify `upgrade-insecure-requests`
- Specify `object-src`: The HTML `object` tag isn't used; restrict it.
- Specify `style-src`: Allow `default-src` to be more restrictive.
- Specify `base-uri`: The HTML `base` tag isn't used; restrict it.
- Include `haptcha.com` subdomains: follows documented requirements.

### DIFF
--- a/bc/settings/project/security.py
+++ b/bc/settings/project/security.py
@@ -2,7 +2,7 @@ import socket
 
 import environ
 
-from ..django import DEVELOPMENT, INSTALLED_APPS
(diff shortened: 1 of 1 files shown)


mlissner:
Looks great. Thank you.

PR #330, opened 2023-07-20
chore(docker): drop ports from ancillary containers

cweider:
`bc2-redis`'s use of `6379` conflicts with `cl-redis` (CourtListener) and frustrates parallel development. There's no need for the `bc2-redis` and `bc2-postgresql` container ports to be exposed on the host. It is enough that they are accessible within the `bc2_net_overlay` Docker network.


mlissner:
Looks good. Thank you.

Issue #350, opened 2023-08-14
Time to disable CSP reporting for Bots.law


mlissner:
@cweider, we got a bunch of false positive CSP reports over the weekend for bots.law. Seems like it's hitting the same thing CourtListener did. Shall we disable the reporting again?

cweider:
Roger – I’ve got a local branch with just that change.

PR #351, opened 2023-08-14
feature(security): drop Content Security Policy violation reporting

cweider:
Since CSP release, we've reached the point where all remaining violations are not caused by Bots.law, but rather by idiosyncratic (and mostly undesirable) browser behavior (largely extensions).

Unfortunately, the volume of these remaining reports is too high; the burden of review is unsustainable. Sacrifice this visibilty and, instead, rely on the regular QA patterns we depend on for other front-end issues (e.g. Javascript exceptions, broken markup).

Fixes: #350


mlissner:
Set for auto merge, thank you!

PR #352, opened 2023-08-16
feat(posts): Add custom border color for groups and remove "Case" from thumbnail

### LINKED ISSUE #344: Custom border color for Little Cases
When we tweet an image like this:

image

The yellow matches the Big Cases bot, not the other bots. I suspect we're going to want to make this configurable.

### LINKED ISSUE #345: Remove "Case:" from thumbnail summaries
When we tweet an image like this:

image

We don't need to say "Case:..." we can just say the name of the case. (Tiny bug.)

If it simplifies the code, we could even remove the header entirely, but I guess it's kind of nice. :)

ERosendo:
This PR fixes #345 and closes #344.

It introduces the following changes:

- Removes the word "Cases" from the thumbnail summaries

- Adds [django-spectrum](https://github.com/dldevinc/django-spectrum) as a dependency.

- Add a field called `border_color` to the `Group` model.

- Add an attr to set the border color in the `TextImage` class.

- Update the logic to use the color from the groups in the templates.

Here are screenshots of the changes in the admin panel and the new thumbnails:

- **New field in the Group form**

     image

     
     Screenshot 2023-08-16 at 8 58 48 AM

- **New thumbnail summaries**


poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry and should not be changed by hand.
```
mlissner, line 1: I think your version of poetry might be old, which explains why this line does not show a version and also why so many lines of the lock file changed. 

Can you make sure it's the latest version and then regenerate the file?

mlissner (approved):

PR #370, opened 2023-08-31
fix(sponsorship): Tweak logic to filter sponsored groups

### LINKED ISSUE #369: AttributeError: 'NoneType' object has no attribute 'pk'
I'm sure what's triggering this. Some corner case having to do with sponsorships, I think? 

It looks like it might be blocking our ability to send updates though...

Sentry Issue: [BIGCASES2-2A](https://freelawproject.sentry.io/issues/4439333141/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'pk'
(2 additional frame(s) were not displayed)
...
  File "bc/subscription/tasks.py", line 316, in process_fetch_webhook_event
    log_purchase(sponsor_groups, subscription.pk, document)
  File "bc/sponsorship/services.py", line 26, in log_purchase
    sponsorship_ids: list[int] = [
  File "bc/sponsorship/services.py", line 27, in <listcomp>
    group.sponsorships.first().pk for group in sponsored_groups if group.sponsorships  # type: ignore
```

ERosendo:
This PR fixes #369.

In line [27](https://github.com/freelawproject/bigcases2/blob/e5a90abf668beeb21d773c9a73e5c12f0aedb7e9/bc/sponsorship/services.py#L27), the code raised an exception and it appears that the list comprehension was unable to obtain the sponsorship PK.

In #369, the transaction we tried to log is related to a document that belongs to a case followed by two sponsored group( big cases and tech cases) but the sponsorship from big cases is not available (we spent all that money :( ). The code should've skipped the big cases group but it tried to get the active sponsorship of that group and that raised the exception. 

I thought that was odd because we're using a filter(`if group.sponsorships`) to check whether the group has a sponsorship or not before trying to get the `pk` and thats when I realized the `.sponsorships` attribute returns a `ManyRelatedManager` object so the `if statement` is always True( the filter is not working 💀 ). I added the .count() method so the if statement will only accepts the items with active sponsorship.

I also added a test for this case.

### DIFF
--- a/bc/channel/selectors.py
+++ b/bc/channel/selectors.py
@@ -55,7 +55,7 @@ def get_sponsored_groups_per_subscription(
     subscription_pk: int,
 ) -> QuerySet[Group]:
     """
-    Returns the set of channel groups that has an active sponsorship and
+    Returns the set of channel groups that has been sponsored and
     are related to the given subscription.
 
     Args:
@@ -70,7 +70,7 @@ def get_sponsored_groups_per_subscription(
             sponsorships__isnull=False,
         )
         .prefetch_related(
-            Prefetch(  # retrieve only Groups related to the subscription
+            Prefetch(  # retrieve only active sponsorships
                 "sponsorships",
                 queryset=Sponsorship.objects.filter(
                     current_amount__gte=3.00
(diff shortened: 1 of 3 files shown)


mlissner (approved):
Looks great. Off it goes!

PR #376, opened 2023-09-15
fix(build): use manual installation for python Poetry 

cweider:
_Analogue of freelawproject/courtlistener#3110_

The Docker image, `python:3.11-slim`, that images are built from, recently stopped working with the installation script that is downloaded and executed for the Poetry tool:
```
RUN curl -sSL https://install.python-poetry.org/ | python3 -
```
Instead of this script, use the installation with `pip` that is a alternative approach which doesn't have this problem and is more direct than the script.

Also in this branch:
- Removal of surplus `POETRY_HOME` argument
- Remove nonoperative `VENV_PATH` environment variable

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -28,15 +28,13 @@ ENV POETRY_VERSION=1.3.2 \
     # do not ask any interactive question
     POETRY_NO_INTERACTION=1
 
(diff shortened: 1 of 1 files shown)


cweider:
Verified run at https://github.com/cweider/bigcases2/actions/runs/6202026845

mlissner:
Lovely, thanks. Merging.

PR #377, opened 2023-09-15
feat(build): eliminate latest-web-dev image

cweider:
_Analogue to freelawproject/courtlistener#3111_

`build` alone is enough for services in Docker compose. Here we remove the `image` configurations and the following bits which become unnecessary:

- `.github/workflows/docker-build-dev.yml` workflow
- development targets of `docker/django/Makefile`


mlissner:
Looks good. Just a warning, nobody is really working on BCB right now, so anything that isn't right may take some time to uncover.

PR #380, opened 2023-09-15
Revert "chore(build): use www-data user for all services"

cweider:
Reverts freelawproject/bigcases2#379

`django-tailwind` does not play well with this change. Addressing this will take some extra attention – revert for now.


mlissner:
Sad trombone, but good to clear it up, thanks.

PR #381, opened 2023-09-16
chore(build): drop use of github archive and merge build-* stages

cweider:
_Analogue to freelawproject/courtlistener#3130_

Building production from a downloaded archive rather than the Docker build context was done to guard against the risk that an unclean working copy could cause WIP code to be deployed.

Today, responsibility for building and deployment are entirely handled by CD process (Github Actions). This is the best practice and has made the archive download safeguard unnecessary.

Change the `build-prod` target to use the Docker buildcontext rather than a downloaded copy of the repository. Once done, the only remaining difference between the two targets is the --without dev option. Eliminate this difference by using `$BUILD_ENV` for the `RUN` instruction and merging the, otherwise identical, stages.

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -53,18 +53,13 @@ FROM build-base as build-prod
 WORKDIR $PYSETUP_PATH
 
 # copy project requirement files here to ensure they will be cached.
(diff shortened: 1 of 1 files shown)


mlissner:
Progress! Thank you!

PR #404, opened 2023-10-17
fix(subscription): Add logic to handle the MultipleObjectsReturned exception

ERosendo:
This PR adds logic to handle the `MultipleObjectsReturned` exception and also adds the `FilingWebhookEvent` model to the admin interface so we can remove duplicate entries.

Here's a screenshot of the new list in the admin panel:

Screenshot 2023-10-17 at 6 35 53 PM

### DIFF
--- a/bc/subscription/admin.py
+++ b/bc/subscription/admin.py
@@ -2,7 +2,7 @@ from django.contrib import admin
 
 from bc.channel.models import Channel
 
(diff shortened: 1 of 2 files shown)


mlissner:
Set for automerge so this gets taken care of, but is there a CL part of this coming soon too?

PR #440, opened 2023-12-15
fix(template): Tweak method to check output validity in the base class

ERosendo:
This PR tweaks the implementation of the `_check_output_validity` method in the `BaseTemplate` class. 

The current implementation of the method in the `BaseTemplate` class is using `len` to count the characters in the output but this approach is inaccurate for Mastodon and Twitter templates because adds the full length of the links to the count. This is not an issue for the Bluesky template class because we implemented a custom method. This issue with the character count led to misleading error messages in the `add-a-case` form and prevented users from following new cases.

This PR adds logic to remove links from the output and fixes the character count. Additionally, it adds a test to prevent regressions.

### DIFF
--- a/bc/core/utils/status/base.py
+++ b/bc/core/utils/status/base.py
@@ -69,13 +69,20 @@ class BaseTemplate:
         """
         Checks whether the provided text exceeds the maximum allowed length.
 
+        Strips links from the output text since they use a fixed character
+        count.
+
         Args:
             text (str): The text to be evaluated.
 
         Returns:
             bool: True if the text length is within the limit, False otherwise.
         """
-        return len(text) <= self.max_characters
+        url_pattern = r"https?://\S+"
+        url_match = re.findall(url_pattern, text)
+        output = re.sub(url_pattern, "", text)
+
+        return len(output) + 23 * len(url_match) <= self.max_characters
 
     def format(self, *args, **kwargs) -> tuple[str, TextImage | None]:
         image = None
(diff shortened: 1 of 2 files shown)


bc/core/utils/status/base.py
```diff
@@ -69,13 +69,20 @@ def _check_output_validity(self, text: str) -> bool:
...
 
         Returns:
             bool: True if the text length is within the limit, False otherwise.
         """
-        return len(text) <= self.max_characters
+        url_pattern = r"https?://\S+"
+        url_match = re.findall(url_pattern, text)
+        output = re.sub(url_pattern, "", text)
+
+        return len(output) + 23 * len(url_match) <= self.max_characters
```
mlissner, line 85: I found it weird that the length of URLs in tweets and toots is both 23, and sort of couldn't believe that this was meant to be in the BaseTemplate. 

Maybe a comment would help:

```suggestion
        url_pattern = r"https?://\S+"
        url_count = len(re.findall(url_pattern, text))
        linkless_output = re.sub(url_pattern, "", text)

        # Twitter and Mastodon both count links as 23 chars at present
        return len(linkless_output) + (23 * url_count) <= self.max_characters
```

(I did some minor refactoring too, mostly just renaming variables.)

mlissner (approved):
Looks good. A couple little things. Merge when ready!

mlissner:
I'm merging this because I need the functionality, but please keep this on your backlog to do another small PR with the two little fixes I suggested.

bc/core/utils/status/base.py
ERosendo: I'll add this comment. Thanks!

PR #471, opened 2024-01-18
349 feat monitor low funding bots

### LINKED ISSUE #349: Monitor when bot funding is getting low
The simplest way to do this is to email the admins when the funding is at around $75. That should give enough time to re-up it, find a new sponsor, etc. 

As it is now, when the funding runs out, nobody notices. 

In future versions, we can probably tweak this to tell the funder and encourage them to re-up, but for now we can keep it simple.

ERosendo:
This PR fixes #349 and introduces the following changes:

- Introduces three new selector functions and adds tests for each of them:
    - `get_curators_by_channel_group_id`
    - `get_total_documents_purchased_by_group_id`
    - `get_groups_with_low_funding`

- Adds a new env variable named `LOW_FUNDING_EMAIL_THRESHOLD`. When funding falls below this threshold, we will start sending notifications to help admin users stay informed and take action.

- Adds an email template for low-funding notifications.

- The `update_sponsorships_current_amount` method now checks the bot's remaining funds after each document purchase. If the remaining funds dip below the established threshold, an email notification is dispatched to ensure timely awareness and action. We use cache to avoid sending the email after each document purchase.

- Adds a Factory class for the Transaction model and updates the tests for the log_purchase function.

- Updates the RQ setting to provide proper configuration for testing and tweaks the `Test` workflow to add a Redis container.


mlissner:
Hmm, I've only skimmed the code so far, but wouldn't it be easier to get the amount remaining for a group before making a transaction, and then if the cost of the purchase pushes it below the threshold, send the email? 

For example: 
 - threshold is $70
 - amount remaining is $71
 - doc costs $3

When we make that purchase, we'd know to send the email, no, and we wouldn't need redis?

bc/sponsorship/emails.py
mlissner, line 1: Here's some suggested text:

### Subject

>  [Action Needed]: Please re-up your funding of {bot_name}

If it's easy, the second message could also say:

> [Action Needed, 2nd Notice]: ...

And the third could say:

> [Action Needed, Final Notice]: ...

(If that's not easy, it's not important.)


### Body

The body could say:

> Hello,
> 
> I am writing because the bot you sponsor on bots.law, {bot_name}, is running low on funding.
> 
> It currently has ${amount_remaining}, and you originally funded it with ${original_amount}.
> 
> Thank you for your sponsorship of this bot! We firmly believe that the support you provide contributes important records to our shared cultural history.
> 
> To continue sponsoring this bot, please do two things:
> 
> 1. Make a payment to Free Law Project at: 
> 
>     https://donate.free.law/forms/pay
> 
> 2. Forward your payment receipt to info@free.law so the money can be allocated to your bot (unfortunately this isn't automated yet!).
> 
> Thank you for your continued support of our bots! Bringing this kind of vital information to everybody is an important way to fight misinformation and educate the public.
> 
> Sincerely,
>
> The Free Law Project Team

bc/sponsorship/utils.py
```diff
@@ -1,22 +1,97 @@
...
+
+    Args:
+        group_id (int): The ID of the channel group to send emails to.
+    """
+    group = Group.objects.get(pk=group_id)
+    curators = get_curators_by_channel_group_id(group_id)
+    total_purchases = get_total_documents_purchased_by_group_id(group_id)
+    curators_emails = list(curators.values_list("email", flat=True))
+
+    email: EmailType = emails["confirm_your_new_account"]
```
mlissner, line 34: This is the wrong key, no?

mlissner (changes requested):
A couple little things, but I think generally we should take the approach that avoids redis. 

You also asked:

> does this mean the intention is to send a single email as a one-time alert(When the remaining amount falls below the set threshold for the first time)? or should we keep sending updates until the amount rises above the threshold?

Could we do it as a list of thresholds? So, send it when it crosses $75, another at $40, and a last one at <$3 (saying that the sponsorship is spent)? 

Maybe the last one could be cc'ed to info@free.law?

bc/sponsorship/utils.py
ERosendo: You're right. This is the wrong key.

ERosendo:
> Could we do it as a list of thresholds? 

Yup. The suggested thresholds seem very reasonable. I'll get them implemented right away.

ERosendo:
@mlissner This PR is ready for another review. 

I've implemented your suggestion about the thresholds and the email's subject. Additionally, I refactored the logic that checks whether we should send an alert or not. Now, it gets the amount remaining for a group before making a transaction.

Let me know what you think.

bc/sponsorship/services.py
```diff
@@ -32,12 +43,40 @@ def log_purchase(
...
+        threshold_value = (
+            settings.LOW_FUNDING_EMAIL_THRESHOLDS[threshold_idx]
+            if threshold_idx is not None
+            else None
+        )
+
+        if (
+            threshold_value
+            and threshold_value > sponsorship.current_amount - document_cost
+        ):
```
mlissner, line 62:
```suggestion
        ):
            # Purchase will cross low funding threshold. Send an email.
```

bc/sponsorship/tests/test_services.py
```diff
@@ -54,6 +65,10 @@ def setUpTestData(cls) -> None:
             page_count=10,
         )
 
+    def tearDown(self) -> None:
+        cache.clear()
```
mlissner, line 69: Do we still need this method?

mlissner (changes requested):
Two little things, please, then feel free to merge if you're ready. We should also do a little manual work to see if we need to contact anybody. Could you do that too after merging?

mlissner (approved):

bc/sponsorship/tests/test_services.py
ERosendo: You're right. We don't need this.

Issue #514, opened 2024-04-18
Fix hashtags in BlueSky posts


mlissner:
They added hashtags, but you have to, like, attach them to the posts, not just put them in the text of the posts, so we need to tweak how we do this. 

Luckily, we applied for a microgrant with BlueSky and got it, so this is actually funded work. Cool! 

We did this for Mastodon in https://github.com/freelawproject/bigcases2/issues/196, if that's helpful.

mlissner:
I'm asking for "micro-applications" from the community to do this work: https://bsky.app/profile/free.law/post/3kqgrb5om4p2b

Woovie:
Hello, I saw your Bluesky post. I do not expect to have time to implement, but here's a some research into the codebase and Bluesky posts in case someone else does:

- https://docs.bsky.app/docs/advanced-guides/post-richtext
- [This seems to be the specific place in which a hashtag facet would be injected, along side the LinkFacet class.](https://github.com/freelawproject/bigcases2/blob/main/bc/channel/utils/connectors/bluesky_api/types.py#L64-L71)
- [Parsing of the links for building facets seems to happen here.](https://github.com/freelawproject/bigcases2/blob/main/bc/channel/utils/connectors/bluesky_api/client.py#L182-L232)

mlissner:
The inventor of hash tags chimes in: https://bsky.app/profile/chrismessina.me/post/3kqixpo5rjk2h

PR #515, opened 2024-04-19
Parse hashtags for Bluesky

mary-ext:
Fixes https://github.com/freelawproject/bigcases2/issues/514

A little rusty with Python, and wasn't sure how I'd get the Bluesky bot to work, thought I'd chip in just a little.

### DIFF
--- a/bc/channel/utils/connectors/bluesky_api/client.py
+++ b/bc/channel/utils/connectors/bluesky_api/client.py
@@ -271,7 +271,7 @@ class BlueskyAPI:
                         "$type": "app.bsky.richtext.facet#tag",
                         "tag": u.text,
                     }
-                ]
+                ],
             }
         return facets
 
--- a/bc/channel/utils/connectors/bluesky_api/client.py
+++ b/bc/channel/utils/connectors/bluesky_api/client.py
@@ -362,7 +362,7 @@ class BlueskyAPI:
 
             for facet in message_object["facets"]:
                 feature = facet["features"][0]
-                if (feature["$type"] == "app.bsky.richtext.facet#link"):
+                if feature["$type"] == "app.bsky.richtext.facet#link":
                     link = feature["uri"]
 
             if link:
--- a/bc/channel/utils/connectors/bluesky_api/client.py
+++ b/bc/channel/utils/connectors/bluesky_api/client.py
@@ -88,6 +88,36 @@ class BlueskyAPI:
         """Get current time in Server Timezone (UTC) and ISO format."""
         return datetime.now(timezone.utc).isoformat()
 
+    def _parse_tags(self, text: str) -> list[RegexMatch]:
+        """
+        Parses hashtags from text.
+
+        This helper function takes a string as input and attempts to extract
+        hashtags from it. If any hashtags are found, they are appended to a
+        list of hashtags. If no hashtags are found, an empty list is returned.
+
+        Args:
+            text (str): The text to parse.
+
+        Returns:
+            list[RegexMatch]: List of matches.
+        """
+        spans = []
+        # reference: https://github.com/bluesky-social/atproto/blob/fbc7e75c402e0c268e7e411353968985eeb4bb06/packages/api/src/rich-text/util.ts#L10
+        # given that our needs of a hashtag is very simple, we can do away with
+        # only parsing alphanumeric characters
+        tag_regex = rb"(?:^|\s)#(?P<tag>[0-9]*[a-zA-Z][a-zA-Z0-9]*)"
+        text_bytes = text.encode("UTF-8")
+        for m in re.finditer(tag_regex, text_bytes):
+            spans.append(
+                RegexMatch(
+                    start=m.start("tag") - 1,
+                    end=m.end("tag"),
+                    text=m.group("tag").decode("UTF-8"),
+                )
+            )
+        return spans
+
     def _parse_urls(self, text: str) -> list[RegexMatch]:
         """
         Parses a URL from text.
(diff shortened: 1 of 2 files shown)


bc/channel/utils/connectors/bluesky_api/client.py
```diff
@@ -229,6 +259,20 @@ def _parse_text_facets(self, text) -> list[TextAnnotation]:
...
+                    "byteStart": u.start,
+                    "byteEnd": u.end,
+                },
+                "features": [
+                    {
+                        "$type": "app.bsky.richtext.facet#tag",
+                        "tag": u.text,
+                    }
+                ],
+            }
```
ERosendo, line 275: This code looks good, but we need to add the `annotation` to the facets list. You can achieve this by appending the `annotation` variable, `facets.append(annotation)`, like line 261

bc/channel/utils/connectors/bluesky_api/client.py
```diff
@@ -313,9 +357,17 @@ def post_status(self, text: str, media: list[Thumbnail]) -> dict[str, str]:
                 "images": media,
             }
         elif message_object["facets"]:
-            card = self.fetch_embed_url_card(
-                message_object["facets"][-1]["features"][0]["uri"]
-            )
+            link: str
+            card: SocialCard | None
```
ERosendo, line 361: Let's assign a default value to these new variables. This will help us avoid the `UnboundLocalError` which can occur if the variable is referenced before it's explicitly assigned a value within the code's scope. Providing a default value ensures the variable is always initialized and has a defined state.

```suggestion
            link: str | None = None
            card: SocialCard | None = None
```

bc/channel/utils/connectors/bluesky_api/client.py
```diff
@@ -88,6 +88,36 @@ def get_current_time_iso(self) -> str:
...
+
+        Returns:
+            list[RegexMatch]: List of matches.
+        """
+        spans = []
+        # reference: https://github.com/bluesky-social/atproto/blob/fbc7e75c402e0c268e7e411353968985eeb4bb06/packages/api/src/rich-text/util.ts#L10
+        # given that our needs of a hashtag is very simple, we can do away with
+        # only parsing alphanumeric characters
+        tag_regex = rb"(?:^|\s)#(?P<tag>[0-9]*[a-zA-Z][a-zA-Z0-9]*)"
+        text_bytes = text.encode("UTF-8")
```
johnludwigm, line 110: If the text is a string, which it appears to be, then there's no need to cast everything to `bytes` objects.
The regex doesn't need the `b` prefix.

mlissner:
Hi @mary-ext and @johnludwigm! Thank you both! 

I'm the director here at Free Law Project, and I'm amazed that this just sort of got done. Open source, man — every once in a while it blows you away. I'll also introduce @ERosendo, who is the maintainer for this area of the code, so that's why he's chiming in with the review. 

@mary-ext, I know you just went ahead and did this work, but if you want to get in touch with me at mike@free.law, I'll be happy to PayPal you  (or whatever) some amount of money for your time on this one (probably should have agreed on that before you jumped in, but I imagine we'll come up with a good number!).

Thank you all!

bc/channel/utils/connectors/bluesky_api/client.py
mary-ext: Looks like I've forgotten to actually append it, thanks!

bc/channel/utils/connectors/bluesky_api/client.py
mary-ext: I would assume that checking if the value isn't None would narrow the type down but Mypy seems to be arguing otherwise, is there something I'm missing on?

image

mary-ext: Ah, dropping the `!= None` works.

mlissner (approved):
LGTM. Let's give it a try!

mlissner:
It works! Thank you @mary-ext!!! 

<blockquote class="bluesky-embed" data-bluesky-uri="at://did:plc:guehcpzih5gpbot5lro2xwy3/app.bsky.feed.post/3kqmijzlrvs2c" data-bluesky-cid="bafyreiaya4tbifm7lrm5krpt57eccovasm3mogxuzvxjc5nwdbdmrbzosu"><p lang="">I&#x27;m now following San Francisco v. Oakland (Frisco/Bay/Oakland airport):[View Full Case]#CL68449160<br><br><a href="https://bsky.app/profile/did:plc:guehcpzih5gpbot5lro2xwy3/post/3kqmijzlrvs2c?ref_src=embed">[image or embed]</a></p>&mdash; Big Cases Bot (<a href="https://bsky.app/profile/did:plc:guehcpzih5gpbot5lro2xwy3?ref_src=embed">@bigcases.bots.law</a>) <a href="https://bsky.app/profile/did:plc:guehcpzih5gpbot5lro2xwy3/post/3kqmijzlrvs2c?ref_src=embed">Apr 20, 2024 at 9:19 PM</a></blockquote>

bc/channel/utils/connectors/bluesky_api/client.py
ERosendo: @johnludwigm Thanks for taking the time to review the PR! I apologize for the late reply.

I'd like to explain the reason behind casting everything to `bytes`: Bluesky strings are internally UTF-8 encoded, and the documentation suggests handling strings in UTF-8 to compute "valid indexes" for elements like Links, Mentions, and Hashtags (you can learn more about it in this [link](https://docs.bsky.app/docs/advanced-guides/post-richtext)). While UTF-8 encoding wouldn't be a problem for plain text posts, we often use emojis in our templates. Emojis can be encoded as multiple characters, leading to incorrect indexes if not handled properly.

I'll create a PR to tweak this helper.

mary-ext: oooh right yeah I forgot about it needing to be in UTF-8 specifically

Issue #461, opened 2024-01-07
Give TwitterConnector, MastodonConnector, et al better repr's


mlissner:
Right now, in Sentry we have a bug that's affecting one of our bots, but it's hard to tell which because the `TwitterConnector` doesn't have a good `__repr__` method. Instead, Sentry shows:

```
<bc.channel.utils.connectors.twitter.TwitterConnector object at 0x7f46584898d0>
```

Eh, we can do better for Twitter and all of our connectors. It doesn't look like we know the twitter account in the object, but we at least know the access_token, which I think is public, right? 

Anyway, this would be nice for debugging and such.

Issue #332, opened 2023-07-25
More enhancements to "I'm now following" posts


mlissner:
I just added a new case for the bot to follow and it dutifully sent out an "I'm now following..." tweet. That's good, but there's a bit of missing context: The bot wasn't following the case until just recently, even though the case was opened over a year ago.

We should add a date stamp to the "I'm now following..." tweets for any case that's older than a month. This should be easy. Just get the date_filed field from the docket object via the API, do a comparison, and then add that bit of context, if needed.

The NYT times does this on its older articles. When you read them, there's a warning at the top that they're old. I think the Guardian does something like this too.

mlissner:
Another change we should do is to include the initial complaint, not just photos of it. So, it should say:

```
Docket: xyz

Complaint: xyz

Context: xyz
```

I think we still have space for that! 

This came up recently when somebody replied to our "I'm now following..." post with a link to the complaint. I was like, what?

TheCleric:
@mlissner looking at this, but I have a question. For something like `enqueue_posts_for_new_case` it's going off the `Subscription` for the case. Do we want to:
1. start saving the `date_filed` there (and possibly backfill it as we go)
1. or just make a call out to CL every time to get the `date_filed`

Option 1 seems like the better one to me, but I'm not sure how you folks feel about messing with database schema (and potentially ugly backfill code).

mlissner:
Yeah, good questions, we could:

 - Pass the date_filed into the method, if we have it.
 - Cal out to CL within the method
 - Add it to the Subscription object

I think tweaking the DB and taking the third option is fine. We don't need to backfill, and it keeps the python code tidy. Works for me. 

@ERosendo, what do you think?

mlissner:
It's already finding the initial complaint and adding it as screenshots on the "I'm now following" posts. I think just use the same document as the one we're using for thumbnails? (Sorry, I forget how that area of the code works, but I think you should have the info you need already in there.)

mlissner:
> As well I'm assuming that you wouldn't want to show the complaint link if this item IS the initial complaint, right?

Most of our "I'm now following" posts are around the time the case comes out. After that, it's in auto-drive mode, and it posts about whatever docs are filed in a given moment. So..I'm not sure I understand your question, sorry!

TheCleric:
Sorry for not being clear: I was referring to this:

> Another change we should do is to include the initial complaint, not just photos of it. So, it should say:

[...]

PR #520, opened 2024-04-27
Tweak some docs links verbiage

TheCleric:
- Updated/added the correct bot links.
- Add a step to the dev instructions that I had to figure out on my own.
- A few verbiage tweaks where they made sense.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -12,7 +12,7 @@ BCB2 uses a [Postgres](https://www.postgresql.org/) database to keep track of ca
 
 ## Posting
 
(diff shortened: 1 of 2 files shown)


TheCleric:
> Thanks! Helpful.
> 
> What are you working on?

I was taking a look at https://github.com/freelawproject/bigcases2/issues/461 as a way of familiarizing myself with the codebase a bit. If there is something particular you feel would be good after that, let me know.

mlissner:
This is a fun one that should be a step harder than what you've done so far, but provide a nice little enhancement:  https://github.com/freelawproject/bigcases2/issues/332

mlissner:
After that, the next priority is a two-fer that should make things more reliable: 

https://github.com/freelawproject/bigcases2/issues/366

And: 

https://github.com/freelawproject/bigcases2/issues/195

Those ones aren't so fun, but they're next in the priority list. After that, the next step is Slack integration (whatever that means!)

PR #521, opened 2024-04-28
feat: Added __repr__ to all connectors + added tests

TheCleric:
Addresses #461 

Also did a little bit of cleanup here. If it's deemed unnecessary let me know and I'll remove it.
- Due to combining the `get_server_url` code and the other mastodon instance splitting code from `self_url`, it is now called `get_handle_parts` and it was moved to the `connectors/mastodon.py` file and tests were updated as well.
- Due to that change I wanted to add tests for `self_url()` to make sure everything worked as expected.
- While doing that it felt proper to have tests for `post_url` as well, which now leverages `self_url` to get its base URL. (Though I wonder if these should be implemented by their connectors instead so we don't have a lot of "business logic" in the channel as we continue to add more channels.

Sorry for the extra rabbit hole of stuff, I got a little carried away. 😆

### DIFF
--- a/bc/channel/models.py
+++ b/bc/channel/models.py
@@ -8,11 +8,7 @@ from bc.users.models import User
 
 from .utils.connectors.base import BaseAPIConnector
 from .utils.connectors.bluesky import BlueskyConnector
-from .utils.connectors.masto import (
-    MastodonConnector,
-    get_server_url,
-    masto_regex,
-)
+from .utils.connectors.masto import MastodonConnector, get_handle_parts
 from .utils.connectors.twitter import TwitterConnector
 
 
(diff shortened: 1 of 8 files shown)


mlissner (approved):
Looks pretty OK to me, but only at a skim. Eduardo, want to do a proper review when you have a moment?

Thanks @TheCleric!

mlissner:
@TheCleric, @ERosendo is assigned to review this, but FYI, he has a couple things in line before this that are higher priority. I expect those may take a few days, but I'd guess you'll get a review this week. 

Thank you!

TheCleric:
Also while I was doing this it's obvious the BlueSky tests are behind Twitter and Mastodon, so I have some extra tests I added that I'll submit in a separate PR after this one is merged.

bc/channel/models.py
```diff
@@ -130,12 +127,10 @@ def self_url(self):
...
             case self.MASTODON:
-                result = masto_regex.search(self.account)
-                assert len(result.groups()) == 2
-                account_part, instance_part = result.groups()
-                return f"https://{instance_part}/@{account_part}"
+                account_part, instance_part = get_handle_parts(self.account)
+                return f"{instance_part}@{account_part}"
             case self.BLUESKY:
-                return f"https://bsky.app/profile/{self.account_id}"
+                return f"https://bsky.app/profile/{self.account}"
```
ERosendo, line 133: Let's use the `account_id` because we store handles without the "@" symbol in there.

ERosendo (approved):
This PR looks good, We can merge it after addressing my comment 👍

mlissner:
I'll duck out. All yours to merge when you're ready, @ERosendo.

bc/channel/models.py
TheCleric: @ERosendo done!

PR #530, opened 2024-05-01
feat(templates): Enhance I'm Now Following [#332]

TheCleric:
This should implement #332 

Things I definitely want others' input on:
- The code for `is_bankruptcy`, which I went with what SEEMED correct, but I could be very wrong (see comment in that function).
- The new templating code that now supports Django templates (see notes below).
- A little bit more strict typing in the CL api code (just to give me some extra insurance that my code was correct).
- I didn't add any new tests to the `test_tasks.py` file. Not sure if that would be good as well, or would overlap a ton with the tests in `test_templates.py` that I added.

For the template code, as I was adding the desired feature, I experimented with a few different ways of doing it. I first attempted to continue our current path of 1 template per option. Since before we only had one option, that was fine since that only meant two templates. However I discovered this grows exponentially. 2 options means 4 templates (e.g., post, post w/ article, post w/ date, post w/ article + date). And 3 options means 8 templates.

This was starting to get unwieldy (especially for the code for deciding which template to use). I also thought about adding something like sub templates where you could inject templates into other templates for some of this conditional stuff. This just felt like writing my own templating engine (poorly). My next experiment was trying to do a lazy loaded f string so I could use f string conditionals in the templates. This proved to be a bit unwieldy as well since to do so I would either have to make the template strings into lambda functions (which seemed funky) or you can populate an f string by passing it into `eval` (which sounds like a security nightmare waiting to happen).

So ultimately the only other choice I could think of boiled down to "use a template language" and since we're using Django, we essentially get one out of the box.

However I know that this may feel like overkill, and others may not share my opinions and evaluations of the other solutions, or even come up with a solution that I didn't think of. So this part is definitely flexible. I just wanted to get something working and in front of you and then iterate from there if needed.


mlissner:
That sounds great, @TheCleric. Let us know when this is ready for review and we'll dig in. Totally makes sense about the template question. The devil will be in the details, but I follow your logic!

TheCleric:
Okay I've finished testing and this seems good to go for me with one big caveat: I don't think I can do the Twitter API without paying money, so that would need someone else to validate.

bc/core/utils/status/templates.py
```diff
@@ -50,23 +45,30 @@
...
+{% if initial_complaint_type and initial_complaint_link %}
+{{initial_complaint_type}}: {{initial_complaint_link}}
+{% endif %}
+
+{% if article_url %}
+Context: {{article_url}}
+{% endif %}
+
+#CL{{docket_id}}""",
 )
```
mlissner, line 72: We deal with this in plain text emails too. I think a more elegant (?) solution is to put your newlines inside your if statements. I'm going to get this wrong, but one approach is something like: 

```suggestion
    str_template="""I'm now following {{docket}}:{% if date_filed %}
Filed: {{date_filed}}{% endif %}
Docket: {{docket_link}}{% if initial_complaint_type and initial_complaint_link %}
{{initial_complaint_type}}: {{initial_complaint_link}}{% endif %}{% if article_url %}
Context: {{article_url}}{% endif %}

#CL{{docket_id}}""",
)
```

TheCleric: Completely agree. That felt like the ugliest bit. I had tried a few ways to avoid it, but never thought of putting the `if`s on the preceding lines!

I have it working now without the `post_replace`. Thanks!

mlissner: Yeah, it's still not pretty, per se, but at least it's self-documenting and hopefully not as fiddly. Thanks for the refactor.

bc/subscription/utils/courtlistener.py
```diff
@@ -308,3 +328,43 @@ def handle_multi_defendant_cases(queue):
         # TODO
         raise NotImplementedError
     logger.debug("handle_multi_defendant_cases(): done")
+
+
+def is_bankruptcy(docket: DocketDict) -> bool:
```
ERosendo, line 333: We can use just the `court_id` as a key identifier for bankruptcy records. We don't need a list of bankruptcy courts. [Juriscraper](https://github.com/freelawproject/juriscraper) employs the following method to identify bankruptcy dockets:

```python
if self.court_id.endswith("b"):
    self.is_bankruptcy = True
else:
    self.is_bankruptcy = False
```

We can simplify it to:

```python
def is_bankruptcy(docket: DocketDict) -> bool:
    ...
    return docket["court_id"].endswith("b")
```

bc/subscription/tasks.py
```diff
@@ -53,11 +58,46 @@ def enqueue_posts_for_new_case(
...
+    days_old = 0
+
+    date_filed_str = docket["date_filed"] if docket else None
+    if date_filed_str:
+        try:
+            date_filed = (
+                datetime.strptime(date_filed_str, "%Y-%m-%d").date()
+                if date_filed_str
+                else None
+            )
```
ERosendo, line 78: the ternary if statement on this line is redundant. We're already checking the same condition on line 72. Could we simplify the code by removing one of them?

bc/subscription/utils/courtlistener.py
ERosendo: After further code review in this PR, I noticed that the `court_id` is already stored in the **Subscription** model. Therefore, for the `is_bankruptcy` method, we can directly pass the `court_id` and leverage the `endswith` method to identify bankruptcy dockets:

```python
def is_bankruptcy(court_id: str) -> bool:
    ...
    return court_id.endswith("b")
```

bc/subscription/tasks.py
```diff
@@ -53,11 +58,46 @@ def enqueue_posts_for_new_case(
...
+    initial_complaint_link = (
+        f"https://www.courtlistener.com{initial_document['absolute_url']}"
+        if initial_document
+        else None
+    )
+
+    initial_complaint_type: Literal["Petition", "Complaint"] | None = None
+    if initial_complaint_link and docket:
+        initial_complaint_type = (
+            "Petition" if is_bankruptcy(docket) else "Complaint"
```
ERosendo, line 95: if we apply the suggestion from https://github.com/freelawproject/bigcases2/pull/530#discussion_r1594384238, we can simplify these lines of code as shown below: 

```python
initial_complaint_type: Literal["Petition", "Complaint"]
initial_complaint_type = (
    "Petition" if is_bankruptcy(subscription.cl_court_id) else "Complaint"
)
```

bc/subscription/tasks.py
```diff
@@ -1,4 +1,5 @@
-from typing import Literal
+from datetime import date, datetime
+from typing import Literal, TypeVar
```
ERosendo, line 2: We're not using the TypeVar import. Let's clean it up.

bc/core/utils/status/base.py
```diff
@@ -1,7 +1,11 @@
 import re
-from dataclasses import dataclass
+from dataclasses import dataclass, field
```
ERosendo, line 2: We can remove the unused `field` import

TheCleric:
@ERosendo thanks for the feedback. I think I've addressed what you pointed out.

ERosendo:
The code looks good and the templates seem well-structured. @mlissner, if you're happy with the content of the templates, I believe we can merge this PR.

Here are screenshots showcasing the different templates:

<details open>
  <summary><h2>New cases</h2></summary>

  - <details>
    <summary><b>Posts with context and initial complaint available for thumbnails</b></summary>

     image


     image


     image

     </details>
- <details>
    <summary><b>Posts with no context and petition available for thumbnails</b></summary>

     image

     image

     image

     </details>

- <details>
    <summary><b>Posts with context and no petition/complaint available for thumbnails</b></summary>

     image

     image

     image
     </details>
</details>


<details open>
  <summary><h2>Old cases</h2></summary>


- <details>
    <summary><b>Posts with context and initial complaint available for thumbnails</b></summary>

     image

     image

     image
     </details>

- <details>
    <summary><b>Posts with no context and petition available for thumbnails</b></summary>
    
     image

     image

     image

     </details>


- <details>
    <summary><b>Posts with context and no petition/complaint available for thumbnails</b></summary>
   
    image

    image

    image

    </details>

</details>

ERosendo (approved):

mlissner:
Looks awesome! This is a huge improvement, thank you for the review Eduardo, and thank you for the great work, @TheCleric. So cool!

PR #531, opened 2024-05-07
feat: Flesh out the BlueSky to tests to the same level as Mastodon and Twitter

TheCleric:
This is a slight followup to #521 

I coded these tests while I was doing that but kept them separate so as not to over-complicate that PR.

These tests are simply the same tests we do for Twitter and Mastodon, just applied to Bluesky so we have test parity.

Also two small typing tweaks were made:
- Change `post_status` to `status_post` in our custom BlueskyAPI to align it with Mastodon and Twitter's API signatures
- Add Bluesky to the `ApiWrapper` type and tweak the `BaseApiConnector`'s types to align with that


mlissner:
Thanks @TheCleric. I'm putting this on @ERosendo's backlog to review!

ERosendo (approved):
Looks good to me! Thanks for the tests, @TheCleric.

PR #533, opened 2024-05-07
feat(bluesky): Handle ReadTimeout exceptions during social card creation

ERosendo:
This PR introduces logic to handle `ReadTimeout` exceptions that may occur during social card creation. When such an exception is encountered, the code will prioritize successfully creating the post itself, ensuring the core functionality remains uninterrupted.

Here's a screenshot of a post with no social card:

image


mlissner (approved):
That'll work!

PR #539, opened 2024-05-09
fix: Add increasing intervals to posting retries (#366 #195)

TheCleric:
This doesn't "fix" #366 or #195, but it may be therapeutic anyway. (See comments in #366).

This will do increasingly delay retries for queued posts in case of failure, with the defaults being 20, 60, and 120 minutes. (Numbers are just a suggestion/guess with no true sound reasoning behind them; feel free to suggest others.)


mlissner:
LGTM. To @ERosendo  for the final review. Sorry Eduardo, lots of PRs your way today!

ERosendo (approved):
LGTM 👍.


Thanks @TheCleric.

PR #552, opened 2024-06-27
feat(deploy): Sync env vars before deployment

blancoramiro:
New step will apply any update to the Bots.law environment variables before running a restart or creating a new pod.

Secret which holds the vars will be updated and then the step will wait for the secret to have a synced status.

Wait has a limit of 30 seconds which if it's reach, the command will exit with an error status.

### DIFF
--- a/.github/workflows/bulid-and-deploy.yml
+++ b/.github/workflows/bulid-and-deploy.yml
@@ -40,8 +40,8 @@ jobs:
       run: aws eks update-kubeconfig --region ${{ env.AWS_REGION }} --name ${{ env.EKS_CLUSTER_NAME }}
     - name: Update Environment Variables
       env:
-        CL_ENV: bots-env
-      run: kubectl annotate es $CL_ENV force-sync=$(date +%s) --overwrite -n ${{ env.EKS_NAMESPACE }} && kubectl wait es -n ${{ env.EKS_NAMESPACE }} --for=jsonpath="{.status.conditions[?(@.reason=='SecretSynced')].status}"=True --timeout=30s $CL_ENV
+        BOTS_ENV: bots-env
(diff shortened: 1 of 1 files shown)


.github/workflows/bulid-and-deploy.yml
```diff
@@ -38,6 +38,10 @@ jobs:
         aws-region: ${{ env.AWS_REGION }}
     - name: Create Kubeconfig with AWS CLI
       run: aws eks update-kubeconfig --region ${{ env.AWS_REGION }} --name ${{ env.EKS_CLUSTER_NAME }}
+    - name: Update Environment Variables
+      env:
+        CL_ENV: bots-env
+      run: kubectl annotate es $CL_ENV force-sync=$(date +%s) --overwrite -n ${{ env.EKS_NAMESPACE }} && kubectl wait es -n ${{ env.EKS_NAMESPACE }} --for=jsonpath="{.status.conditions[?(@.reason=='SecretSynced')].status}"=True --timeout=30s $CL_ENV
```
mlissner, line 44:
```suggestion
        BOTS_ENV: bots-env
      run: kubectl annotate es $BOTS_ENV force-sync=$(date +%s) --overwrite -n ${{ env.EKS_NAMESPACE }} && kubectl wait es -n ${{ env.EKS_NAMESPACE }} --for=jsonpath="{.status.conditions[?(@.reason=='SecretSynced')].status}"=True --timeout=30s $BOTS_ENV
```

mlissner (approved):
One suggestion, similar to the one I made earlier today. :)

.github/workflows/bulid-and-deploy.yml
blancoramiro: Nice catch, thank you. Updated!

PR #558, opened 2024-07-05
Replace logos

quevon24:
- Replace normal and responsive logo in header (free.law clickable)
image
image

- Replace FLP logo
image
image

- OG tags and image added
image

Work in progress....


mlissner:
Man, having the logo inset in the little square sure doesn't look great. I wonder if we'll need to do it more like on Free.law? 

If we do that, we might need to change the whole color scheme of the navbar, but that's not horrible if it fixes things.

Another idea: A yellow border on the inset?

quevon24:
> Man, having the logo inset in the little square sure doesn't look great. I wonder if we'll need to do it more like on Free.law?
> 
> If we do that, we might need to change the whole color scheme of the navbar, but that's not horrible if it fixes things.
> 
> Another idea: A yellow border on the inset?

mmm maybe we can make the logo bigger, add a padding and use a gray border like the one used in the header, somethinglike this: 
image

mlissner:
1000× better! :shipit:

quevon24:
@mlissner Which FLP logo looks best? black or purple? (i updated the images above)

mlissner:
Purple in this context, please.

quevon24:
@mlissner about the favicon, i think  the scale has too many details and it looks quite bad:
image

with white bg:
image

mlissner:
Jenifer is working on a simplified version that we hope will work.

quevon24:
> Jenifer is working on a simplified version that we hope will work.

great, because I didn't really like the result I got

mlissner:
I'd suggest doing your PR without favicons and adding them in a later one.

quevon24:
@mlissner The remaining changes are the favicon, and the robot that appears in the footer, in the spreadsheet it is not clear if it needs to be replaced or not.

this: 
image

I think it looks good, so the site won't be so loaded with logos

Maybe the sticker or something similar would look good, it's a similar color to the robot: https://github.com/freelawproject/logos/blob/1480214df48e664fec37430ac56408dc962a9fe8/sites/bots.law/v1/stickers/round-color.svg

bc/web/templates/includes/logo.svg
mlissner, line 1: Can you make the rectangle smaller, so it only covers the "by Free.law" part more accurately?

bc/assets/templates/includes/header.html
```diff
@@ -4,9 +4,12 @@
     <div class="max-w-7xl mx-auto px-4 sm:px-6 md:px-10">
         <div class="flex w-full justify-between items-center py-6 md:space-x-6 lg:space-x-10">
             <div class="absolute top-6">
-                <a href="{% url 'homepage' %}">
-                    <img alt='Yellow robot with white background logo' src="{% static "svg/icon.svg" %}"  layout="responsive" class="rounded-lg border-4 border-bcb-black h-20 w-[80px] lg:h-24 lg:w-24 bg-white"/>
-                </a>
+                    <div alt='Yellow robot with white background logo'  layout="responsive" class="rounded-lg border-2 p-2 border-gray-400 h-30 w-[120px] lg:h-32 lg:w-44 bg-white">
```
mlissner, line 7: This `alt` is no longer correct. Can you update it, please? 

Also, can you see if it's possible to avoid using the `px` classes and prefer using the other ones, with numbers, like `w-5` instead of `w-[40px]` or whatever? I'm not sure why, but I have a pretty strong feeling that using the px-based ones is bad form. Maybe somebody else knows? (You use these down below too.)

mlissner (changes requested):
Looks great. Just a couple little things.

bc/assets/templates/includes/header.html
quevon24: The documentation comments that you can use the px values ​​to perfectly position the element: https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values

I can use the other classes but it is not perfect, I used px adjustment so that when opening the responsive menu the logo is seen almost above the header logo

mlissner (approved):

mlissner:
Great, thank you Kevin!

PR #561, opened 2024-07-09
Add new icons to bots.law

mlissner:
This follows the pattern set in https://github.com/freelawproject/courtlistener/pull/4184 to create new favicons and manifest icons. In general, these are the same icons as in that PR, but tweaked to the bots.law context and colors. Again, no dark them variants were needed, thankfully. 

This also reduces the indentation of base.html so it's two spaces.


ARCHITECTURE.md
```diff
@@ -38,7 +38,7 @@ To set up a development machine, do the following:
 
 1. `cd` into bigcases2/docker/bigcasesbot, then launch the server by running:
 
-    `docker-compose up -d`
+    `docker compose up --build`
```
quevon24, line 41:
```suggestion
    `docker compose up -d --build`
```
I think it is better to run it in detached mode and use tools like docker desktop to monitor the service logs

quevon24 (approved):
All good, just a little suggestion

PR #562, opened 2024-07-09
feat(imgs): Update favicons

mlissner:


flooie (approved):
Looks great.  I really like the rounded corners and think that will look good across browsers

PR #579, opened 2024-08-22
feat(subscription): Optimize document handling in subscription tasks

ERosendo:
This pull request addresses the memory-related issues identified in #577 by significantly enhancing the task enqueueing code. 

Key improvements:

- Refactored helper methods: The helper method responsible for enqueuing tasks has been updated to send only document URLs instead of entire byte objects. This substantially reduces memory usage, particularly for large documents.
- The task code has been updated to handle document URLs effectively.

### DIFF
--- a/bc/subscription/tasks.py
+++ b/bc/subscription/tasks.py
@@ -39,7 +39,7 @@ queue: Queue = get_queue("default")
 
 def enqueue_posts_for_new_case(
     subscription: Subscription,
-    document: bytes | None = None,
+    document: bytes | str | None = None,
     check_sponsor_message: bool = False,
     initial_document: DocumentDict | None = None,
 ) -> None:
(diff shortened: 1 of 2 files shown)


bc/subscription/tasks.py
```diff
@@ -341,7 +352,7 @@ def process_fetch_webhook_event(
             "The RECAP document doesn't have a path to download the file"
         )
 
-    pdf_data = download_pdf_from_cl(cl_document["filepath_local"])
+    pdf_data = cl_document["filepath_local"]
```
mlissner, line 355: Probably we should rename the variable here?

mlissner (approved):
Cool. One little suggestion, but it can wait an hour for this to be merged and deployed and for the next PR to do it. :)

PR #608, opened 2024-09-24
feat(utils): Upgrade to CourtListener v4 API

### LINKED ISSUE #607: Upgrade bots.law to v4 of the API
We should probably eat our own dogfood and switch Bots.law to use v4 of the API. 

The migration guide is here: 

https://www.courtlistener.com/help/api/rest/v4/migration-guide/

I suspect this will be very easy, but if the guide isn't good enough or if the APIs fall short, that's bad and we should think about ways to fix the docs or the API.

ERosendo:
This PR fixes #607 by replacing all API calls to CourtListener with the latest v4 version, ensuring compatibility with the newest features and improvements.


mlissner (approved):
Simple enough, thank you!

PR #609, opened 2024-09-24
feat(auth): Uses Ada to redirect users upon login

ERosendo:
This PR adds Ada as a dependency and creates a custom `LoginView` class that validates and sanitizes the redirect URL after a successful login.

Key changes in this PR:

- Adds https://github.com/ada-url/ada-python as a dependency

- Adds a new helper method to parse and validate the redirect URLs using Ada

- Introduces a custom `LoginView` class called `SafeRedirectLoginView` that inherits from Django's built-in LoginView but adds an extra layer of security by ensuring the redirect URL submitted by the login form is safe.


mlissner:
Looks good, thanks.

PR #610, opened 2024-09-24
fix(auth): Set cache control header in authenticated pages

### LINKED ISSUE #599: Issue with cache control
This report highlights a potential issue with the website's cache control headers. Improper cache settings may result in users viewing outdated content or experiencing unexpected behavior.

Steps to Reproduce:

- Visit the login page: https://bots.law/sign-in/
- Enter valid credentials and successfully log in to your account.
- Navigate to the profile settings page: https://bots.law/profile/settings/
- Click the logout button to exit your account and return to the login page.
- Click the browser's back button.
- Instead of being redirected to the login page to re-enter credentials, you get into account, but with limited access (read-only).

ERosendo:
This commit fixes #599 by setting the cache control header to all views that require a logged-in user (using the [never_cache()](https://docs.djangoproject.com/en/5.1/topics/http/decorators/#django.views.decorators.cache.never_cache) decorator), ensuring that the content is not cached by browsers and is always fetched from the server.


mlissner:
Nice one, thanks. Are we fixing this on CL too?

ERosendo:
>  Are we fixing this on CL too?

@mlissner CL is not experiencing this caching problem. Most authenticated views are already using the `never_cache` decorator.

PR #618, opened 2024-10-31
feat(channels): Add support for Threads

elisa-a-v:
Addresses [this issue](https://github.com/freelawproject/foresight/issues/13)

This PR introduces full integration with the Threads API, allowing the bots to post directly to Threads.

### Considerations

1. The official Threads API requires images to be hosted on a public server, so we need to upload them to S3 first. To achieve this:
	- Set up a bucket with the proper policy, which should at least have the statements shown below, replacing `<BUCKET_NAME>`, `<AWS_ACCOUNT_ID>`, and `<USER>` with the username for the IAM user whose credentials are specified in the AWS `.env` variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).
	- Add the bucket name and region to your environment variables as `THREADS_BUCKET_NAME` and `THREADS_BUCKET_ZONE`.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::<AWS_ACCOUNT_ID>:user/<USER>"
            },
            "Action": [
                "s3:PutObject",
                "s3:GetObject",
                "s3:PutObjectAcl"
            ],
            "Resource": "arn:aws:s3:::<BUCKET_NAME>/*"
        },
        {
            "Sid": "PublicReadGetObject",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::<BUCKET_NAME>/*"
        }
    ]
}
```

2. To generate the access tokens for the bots:
	- You need a meta developer account to create an app with the use case "**Access to Threads API**"
	- Enable the permissions `threads_basic` and `threads_content_publish` in your app.
	- Add a `Threads Tester` Role with the Threads user for the bot.
	- Add the `THREADS_APP_ID` and `THREADS_APP_SECRET` to the `.env` file
	- Run the script `scripts/get-threads-keys.py` and follow the instructions which will provide the credentials to create a Channel via Django admin.

### TO DO:
- Configure bucket to delete images after a set period of time. They don't need to be kept for long, we only need them until we have created the container in Threads.
- Add instructions to wiki.
- Tests?


.env.example
```diff
@@ -47,8 +47,6 @@ SENTRY_REPORT_URI=""
 # twitter.py
 TWITTER_CONSUMER_KEY=""
 TWITTER_CONSUMER_SECRET=""
-TWITTER_ACCESS_TOKEN=""
-TWITTER_ACCESS_TOKEN_SECRET=""
```
elisa-a-v, line 51: These were removed because they weren't used anymore, as discussed with @ERosendo

mlissner (approved):
I gave it a quick skim. Looks about right, generally!

One thing to consider is the png->jpeg code we discussed today. My instinct on the call was that Doctor should be producing jpeg thumbnails for you instead of pngs, but jpeg is an inferior format for this and I wouldn't want to post jpegs on BlueSky or Twitter, where PNG makes more sense. So I think that means you could:

 - Call doctor to get png thumbnails for those, then...
 - Call doctor to get jpegs for threads

Or:
 - Call doctor for the pngs, then...
 - Convert them here.

Honestly, the latter seems easier and you've got the code. :shipit: !

I'm also a bit curious about your use of envs, but I'll let Eduardo chime in on that if we need changes there.

Finally, can you please open an issue to update our website to put the threads info on it in all the right places?

bc/channel/utils/connectors/threads.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Uploads media to public storage for Threads API compatibility.
+
+        Since Threads API requires media to be accessible via public URL,
+        this method resizes the image as needed, uploads it to S3, and
+        returns the public URL.
+
+        Args:
+            media (bytes): The image bytes to be uploaded.
+            _alt_text (str, optional): Alternative text for accessibility
```
ERosendo, line 45: I found we can include alt text for images when using the API – that's great for accessibility. Here's the documentation for it on Meta's developer site:

https://developers.facebook.com/docs/threads/posts/accessibility

bc/channel/utils/connectors/threads_api/client.py
```diff
@@ -0,0 +1,174 @@
...
+from bc.core.utils.images import convert_to_jpeg, resize_image
+from bc.core.utils.s3 import put_object_in_bucket
+
+logger = logging.getLogger(__name__)
+
+
+_BASE_API_URL = "https://graph.threads.net/v1.0"
+
+
+class ThreadsAPI:
```
ERosendo, line 17: All get and post requests must have timeouts: https://github.com/psf/requests/issues/3070

bc/channel/models.py
```diff
@@ -131,6 +138,8 @@ def self_url(self):
                 return f"{instance_part}@{account_part}"
             case self.BLUESKY:
                 return f"https://bsky.app/profile/{self.account_id}"
+            case self.THREADS:
+                return f"https://www.threads.net/@{self.account_id}"
```
ERosendo, line 142: Let's use `account` instead of `account_id`. After generating my token using the `get-threads-keys.py` script, I noticed the suggested value for `account_id` is simply a number.

```suggestion
                return f"https://www.threads.net/@{self.account}"
```

bc/settings/third_party/threads.py
```diff
@@ -0,0 +1,9 @@
+import environ
+
+env = environ.FileAwareEnv()
+
+THREADS_APP_ID = env("THREADS_APP_ID", default="")
+THREADS_APP_SECRET = env("THREADS_APP_SECRET", default="")
+THREADS_CALLBACK_URL = env("THREADS_CALLBACK_URL", default="")
+THREADS_BUCKET_NAME = env("THREADS_BUCKET_NAME", default="")
+THREADS_BUCKET_ZONE = env("THREADS_BUCKET_ZONE", default="")
```
ERosendo, line 9: Based on the default value in `put_object_in_bucket`, let's rename this setting to `THREADS_BUCKET_REGION`. (We should also update the .env.example file to reflect this change)

ERosendo (changes requested):
@elisa-a-v The code looks great!  Just wanted to point out a couple of things we might have missed when creating this [to-do list](https://github.com/freelawproject/foresight/issues/13#issuecomment-2430270756):

1. **Threads logo**: We need to add an `SVG` file for the Threads logo (the one we use in buttons). These files usually go in the `inline` folder inside the `includes` directory of the `assets` folder.

2. **Follow button template**: The `follow-button.html` template needs to be updated to use the `SVG` file mentioned above. This fragment is used in the detail page for the little cases bot. For reference, I've included the red rendered by this template below:

     Screenshot 2024-11-04 at 1 30 24 PM

     We need the logo to make the button look complete:

    image

3. **AWS Credentials**: I encountered some issues while setting up the AWS credentials. This might be a good opportunity to revisit our approach and potentially align it with the method used for courtlistener. Here's the relevant issue in courtlistener for reference: https://github.com/freelawproject/courtlistener/issues/4018

4. **Token Refresh Mechanism**: This PR introduces a script to generate our initial long-lived tokens. However, it currently lacks a mechanism for automatic token refresh before expiration. 

    We can merge this PR as-is and create a separate issue to address the token refresh functionality. However, We should prioritize and implement this mechanism before our tokens expire.

bc/settings/third_party/threads.py
elisa-a-v: [Done](https://github.com/freelawproject/bigcases2/pull/618/commits/f171af43be509313cec0fd7e6eb5a26dc6b9ec89)!

ERosendo (approved):
LGTM :shipit:

mlissner:
Exciting. Shall I merge and what's the best strategy for launching?

elisa-a-v:
> Shall I merge

Sure! If you both approved, go for it.

> and what's the best strategy for launching?

Well, I'm not sure what you mean, I guess the first step would be to create a big cases Thread account if there isn't one, then run the script and create the new channel.

Oh and btw, you asked me to open an issue to add the Threads bot info to the website and everywhere else it should be displayed, should I open that issue in this same repo or a different one?

mlissner:
Merged! Here we go, thank you! 

For launching, I do mean how we create the accounts, set them up, etc, all that fun stuff. I'm happy to help if needed, but ideally we'd have this documented on the wiki, and it'd be something I wouldn't need to be too involved in? Maybe Gianfranco can help instead since he has server access? 

Yep, the issue for adding threads info the website should go here. :)

ERosendo:
@mlissner Elisa created a new wiki page that outlines the steps needed to create a Threads app and get values for new environment variables.


https://github.com/freelawproject/bigcases2/wiki/One%E2%80%90Time-Setup-to-enable-Threads

Issue #646, opened 2024-12-10
Invalidate cache in the Banner Config save method


elisa-a-v:
#643 introduced a customizable banner with its content managed from a DB model. Since the bots.law pages are cached in the CDN, changes in the DB aren't always shown in all locations, so we need to invalidate the cache in the `BannerConfig`'s `save` method for the changes to propagate accordingly.

PR #647, opened 2024-12-24
feat(core): Invalidate cache after saving banners

ERosendo:
This PR addresses #646 by using the post-save signal to invalidate the cache whenever a BannerConfig object is saved or updated.

### DIFF
--- a/bc/core/apps.py
+++ b/bc/core/apps.py
@@ -1,0 +1,9 @@
+from django.apps import AppConfig
+
+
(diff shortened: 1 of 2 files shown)


mlissner (approved):
LGTM. @elisa-a-v, want to do the final review so you can see the approach and merge?

elisa-a-v (approved):
Yes thank you! it looks good so I'll go ahead and merge

PR #703, opened 2025-05-06
chore: Move formatting and linting to Ruff

adamchainz:
Replace Black, isort, pylint, and pyupgrade with Ruff.

I did not add the `project` metadata in `pyproject.toml` for this repository because Poetry reports a conflict with its `tool.poetry` metadata. I'll add `project` only when converting to uv.

### DIFF
--- a/bc/channel/apps.py
+++ b/bc/channel/apps.py
@@ -6,4 +6,4 @@ class UsersConfig(AppConfig):
 
     def ready(self):
         # Implicitly connect a signal handlers decorated with @receiver.
-        from bc.channel import signals
+        from bc.channel import signals  # noqa: F401
--- a/bc/channel/models.py
+++ b/bc/channel/models.py
@@ -6,10 +6,10 @@ from redis.exceptions import LockError
 
 from bc.core.models import AbstractDateTimeModel
 from bc.core.utils.color import format_color_str
+from bc.core.utils.redis import make_redis_interface
 from bc.sponsorship.models import Sponsorship
 from bc.users.models import User
 
-from ..core.utils.redis import make_redis_interface
 from .utils.connectors.base import (
     BaseAPIConnector,
     RefreshableBaseAPIConnector,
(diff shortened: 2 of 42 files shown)


bc/channel/apps.py
```diff
@@ -6,4 +6,4 @@ class UsersConfig(AppConfig):
 
     def ready(self):
         # Implicitly connect a signal handlers decorated with @receiver.
-        from bc.channel import signals
+        from bc.channel import signals  # noqa: F401
```
adamchainz, line 9: Ruff's [F401 / unused-import rule](https://docs.astral.sh/ruff/rules/unused-import/) removes unused imports with the `--fix` option. This is generally desirable, but in the case of such implicit-action imports, we need to ignore the rule.

bc/channel/models.py
```diff
@@ -182,12 +181,13 @@ def validate_access_token(self):
...
-                lock.release()
-            except Exception as e:
-                logger.error(f"Error releasing lock for channel {self}:\n{e}")
+            if lock.owned():
+                try:
+                    lock.release()
+                except Exception as e:
+                    logger.error(
+                        f"Error releasing lock for channel {self}:\n{e}"
+                    )
```
adamchainz, line 190:
```
bc/channel/models.py:186:17: B012 `return` inside `finally` blocks cause exceptions to be silenced
    |
184 |         finally:
185 |             if not lock.owned():
186 |                 return
    |                 ^^^^^^ B012
187 |             try:
188 |                 lock.release()
    |
```

https://docs.astral.sh/ruff/rules/jump-statement-in-finally/

This construct is actually deprecated from Python 3.14 due to its error-inducing potential: https://docs.python.org/3.14/whatsnew/3.14.html#pep-765-disallow-return-break-continue-that-exit-a-finally-block 

The fix here avoids the `return` by moving the `lock.release()` call under the `if lock.owned()` condition

albertisfu (commented):
Thanks @adamchainz this looks good.

I just noticed that the old formatting/linting dependencies, Black, Flake8, isort, etc. are still listed under dev-dependencies in `pyproject.toml.` Is that also related to your comment?

> I did not add the project metadata in pyproject.toml for this repository because Poetry reports a conflict with its tool.poetry metadata. I'll add project only when converting to uv.

Also, those dependencies are still present in `dependabot.yml`.

albertisfu (approved):

mlissner:
Got it and removed it. Thank you both! Merged!

Issue #617, opened 2024-10-31
Punctuation gets mangled


mlissner:
In [this post](https://bsky.app/profile/bigcases.bots.law/post/3l7t5p6qbbm2u), I summarized it as "Elon's election interference" and on BlueSky we got:

![Image](https://github.com/user-attachments/assets/3bb02904-e1e9-4db7-9c0e-188b297fcee5)

Same on Twitter too:

![Image](https://github.com/user-attachments/assets/9b4f1d9c-a288-4fe6-abbd-b6fcc5e1660e)

ajonnav:
Hi, this is my first time working on an issue so let me know if I am doing things wrong.

I was able to reproduce this issue -- the problem seems to be that Django templates autoescape HTML characters. One potential fix would be to add the `autoescape` tag (https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#autoescape) to the templates; another similar fix would be to add the `safe` filter (https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#safe) to variables that do not need to be escaped. It seems like the first option would be easier to implement and perhaps a little more maintainable -- I would just enclose the existing templates in the `autoescape` tag, and people would just need to ensure that they do the same for new templates.

Let me know if you have concerns. In particular, it is not clear to me where the input values for the templates come from so let me know if not escaping the inputs would be a bad idea.

mlissner:
Hi! Thanks for picking up this issue and for stepping forward to claim it. You're doing this the right way. 

I'm not in this code enough to have an opinion about this, but the data is generally considered safe (it's coming exclusively from trusted curators). Want to do a little PR and see how that goes?

PR #706, opened 2025-05-06
Do not escape variables in follow-new-case templates

ajonnav:
Fix for #617. Apostrophes and other punctuation for Django templates were getting mangled because Django autoescapes HTML characters. This PR fixes the issue by enclosing the Big Cases Bot templates that are Django templates—the Follow-New-Case templates—within the Django `autoescape` tag. The tests added failed without the code change, and passed after the code change.


mlissner:
Thanks @ajonnav! We handle PRs in either a slow or fast queue, depending on priority. Items in the fast queue are allowed to interfere with our current sprints and items in the slow queue are schedule for the next sprint. This is a slow-queue item, so it will be reviewed during our next sprint, which will start in about two weeks. Sorry for the delay on this, but rest assured that we will get to it!

ERosendo:
Thanks @ajonnav. LGTM :shipit:

